PR #31, opened 2020-07-21
Redirects user to lecture page and grabs lecture key from URL

jaysaleh:
This pull request redirects each lecture to the `lecture-view.html` and grabs the key from the URL.


src/main/webapp/lectures.js
```diff
@@ -1,3 +1,7 @@
+/* Used to redirect users to the lecture view page, while storing lecture key in
+ * the URL. */
+const LINK_PREFIX = '/lecture-view.html?key=';
```
n-wach, line 3: I'd recommend using `URL.searchParams` to construct this.  For instance:
```
  const url = new URL('/lecture-view.html', window.location.origin);
  url.searchParams.append('key', key);
```

src/main/webapp/lectures.js
```diff
@@ -23,8 +27,7 @@ function createLectureListItem(lecture) {
   const listItem = document.createElement('li');
   const lectureLink = document.createElement('a');
 
-  // TODO: Redirect to specific lecture page.
-  lectureLink.href = lecture.videoUrl;
+  lectureLink.href = LINK_PREFIX + lecture.key.id;
```
n-wach, line 30: Use `URL` object, see above

src/main/webapp/view.js
```diff
@@ -0,0 +1,15 @@
+/* Number of characters until lecture ID in URL. */
+const URL_ID_SPACING = 5;
```
n-wach, line 2: Where does the 5 come from? `?key=`?

n-wach: I think we should just use URL and URLSearchParams to do this parsing/generating

jaysaleh: i agree, ill make the change

src/main/webapp/lectures.js
jaysaleh: fixed

src/main/webapp/lectures.js
jaysaleh: fixed

PR #40, opened 2020-07-21
Added URL builder to LectureServlet.java

jaysaleh:
This pull request adds support for redirecting to new lecture page after user enters a new lecture.

### DIFF
--- a/src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
+++ b/src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
@@ -71,7 +71,7 @@ public class LectureServlet extends HttpServlet {
     Entity lectureEntity = new Entity(Lecture.ENTITY_KIND);
     lectureEntity.setProperty(Lecture.PROP_NAME, lectureName);
     lectureEntity.setProperty(Lecture.PROP_URL, videoUrl);
-    lectureEntity.setProperty(Lecture.PROP_ID, videoId);
+    lectureEntity.setProperty(Lecture.PROP_VIDEO_ID, videoId);
 
     datastore.put(lectureEntity);
     response.sendRedirect(buildRedirectUrl(lectureEntity));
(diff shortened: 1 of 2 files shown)


src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -69,15 +73,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     lectureEntity.setProperty(Lecture.PROP_URL, videoUrl);
     lectureEntity.setProperty(Lecture.PROP_ID, videoId);
 
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     datastore.put(lectureEntity);
-    response.sendRedirect(REDIRECT_URL);
+    response.sendRedirect(
+        buildRedirectUrl(String.valueOf(lectureEntity.getKey().getId()), videoId));
```
n-wach, line 78: Maybe have buildRedirectUrl take an entity or Key instead of string? Because I assume you'll never be passing something other than an lecture's key

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -112,4 +114,22 @@ private String getVideoId(String videoUrl) {
...
+   * found in {@code lectureEntity}.
+   */
+  private String buildRedirectUrl(Entity lectureEntity) {
+    String lectureId = String.valueOf(lectureEntity.getKey().getId());
+    String videoId = (String) lectureEntity.getProperty(Lecture.PROP_ID);
+
+    try {
+      URIBuilder urlBuilder = new URIBuilder(REDIRECT_URL)
+                                  .addParameter("id", lectureId)
+                                  .addParameter("videoId", videoId);
```
n-wach, line 129: Should these params be constants?

PR #78, opened 2020-07-28
Use Mockito for testing

n-wach:
This removes our dependency on spring-mock and demos using Mockito for testing with httpservletrequest and response


src/test/java/com/googleinterns/zoomtube/mocks/MockResponse.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.mockito.Mockito.mock;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
+
+/**
+ * This mocks {@link javax.servlet.http.HttpServletResponse} to allow for testing
+ * servlets. As more tests and servlets are added, you may need to override additional
```
amuamushu, line 12: nit: Having "you" in the javaDoc is a bit weird to me personally. Can you rephrase it to say something like 
"As more tests and servlets are added, additional methods may need to be overridden."? (This also shows up in MockRequest)

src/test/java/com/googleinterns/zoomtube/mocks/MockRequest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+
+/**
+ * This mocks {@link javax.servlet.http.HttpServletRequest} to allow for testing
+ * servlets. As more tests and servlets are added, you may need to override additional
+ * methods.
+ */
+public class MockRequest extends HttpServletRequestWrapper {
```
f-camp, line 18: This is more of a fake than a mock at this point and passing in the mock servlet request in the wrapper is redundant as you are implementing the other methods now..

I noticed that HttpServletRequest is an interface. You should be able to make this a fake without Mockito then.
```public class MockRequest implements HttpServletRequest``` will do this for you. And then people can use your fake requests every where and implement methods as needed.

File: src/test/java/com/googleinterns/zoomtube/mocks/MockResponse.java [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+ * methods.
+ */
+public class MockResponse extends HttpServletResponseWrapper {
```
f-camp, line 15: Ditto.

src/test/java/com/googleinterns/zoomtube/mocks/MockRequest.java
n-wach: The problem is that the interface has dozens of methods which all need to be implemented.  The Wrapper class wraps an existing implemention so you don't need to reimplement them all.  However, you need to provide an underlying instance, and using mock let's me instantiate one.  I could probably use null instead.  I just want to avoid having to write out every method of the interface--there are a lot

f-camp: You can leave the methods we don't need unimplemented and empty and just implement the ones we actually need. 
Sending a mock in the wrapper class could have other unwanted consequences as well as the  class could be calling methods for which you don't have when...then clauses configured for it. Having a mock in a wrapper class you do not own can become unwieldy to debug. 

Seems to me the two options here are to have the when..then.. clauses in the test itself and mock out HttpServletRequest there or continue with the Fake here but implement it properly. 
I understand you would want the utility methods of ```contentAsString()``` so it sounds like you want the proper Fake?

src/test/java/com/googleinterns/zoomtube/servlets/AuthenticationServletTest.java
```diff
@@ -9,25 +11,30 @@
...
   private AuthenticationServlet servlet;
-  private MockHttpServletRequest request;
-  private MockHttpServletResponse response;
+  @Mock private HttpServletRequest request;
+  @Mock private HttpServletResponse response;
   private LocalServiceTestHelper authService;
 
   @Before
-  public void setUp() throws ServletException {
+  public void setUp() throws ServletException, IOException {
```
f-camp, line 35: Feel free to throws Exception in tests

n-wach: Changed

src/test/java/com/googleinterns/zoomtube/servlets/AuthenticationServletTest.java
```diff
@@ -38,77 +45,92 @@ public void cleanUp() {
...
 
     servlet.doGet(request, response);
 
-    assertThat(response.getContentType()).isEqualTo("application/json;");
-    String json = response.getContentAsString();
+    verify(response).setContentType("application/json;");
+    String json = content.toString();
     Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
     AuthenticationStatus status = gson.fromJson(json, AuthenticationStatus.class);
     assertThat(status.loggedIn()).isFalse();
```
amuamushu, line 79: nit: can you separate the assert?

n-wach: The `verify` and json stuff is part of the assertions/test section

PR #81, opened 2020-07-29
Add tests for discussion servlet

n-wach:


n-wach:
Oops, need to wait for #78 to be merged to shrink changes

n-wach:
Nevermind, I can just change the base branch to `mockito`.

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -47,8 +47,8 @@
 // TODO: References to Lecture kind string literal need to be updated once Lecture code is merged.
 @WebServlet("/discussion")
 public class DiscussionServlet extends HttpServlet {
-  private static final String PARAM_LECTURE = "lecture";
-  private static final String PARAM_PARENT = "parent";
+  public static final String PARAM_LECTURE = "lecture";
```
jaysaleh, line 50: Does @visiblefortesting work here?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    Type listType = new TypeToken<ArrayList<Comment>>() {}.getType();
+    List<Comment> comments = gson.fromJson(json, listType);
+    assertThat(comments.size()).isEqualTo(2);
+  }
+
+  private Entity createTestCommentEntity(int lectureId) {
+    Entity commentEntity = new Entity(Comment.ENTITY_KIND);
+    Key lecture = KeyFactory.createKey(/* kind= */ "Lecture", lectureId);
+    commentEntity.setProperty(Comment.PROP_LECTURE, lecture);
```
jaysaleh, line 167: Maybe we should create a util class for Comment Entity consts rather than having them in the AutoValue class. What do you think?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    assertThat(comments.size()).isEqualTo(3);
+  }
+
+  @Test
+  public void doGet_returnsCommentsForSpecificLecture() throws ServletException, IOException {
+    request.setParameter(DiscussionServlet.PARAM_LECTURE, "2");
+    datastore.put(createTestCommentEntity(1));
+    datastore.put(createTestCommentEntity(1));
+    datastore.put(createTestCommentEntity(1));
+    datastore.put(createTestCommentEntity(2)); // this one
```
jaysaleh, line 149: These comments are confusing

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  @Test
+  public void doGet_returnsNothingForUnknownLecture() throws ServletException, IOException {
+    request.setParameter(DiscussionServlet.PARAM_LECTURE, "1");
+
+    servlet.doGet(request, response);
+
+    assertThat(response.getContentType()).isEqualTo("application/json;");
+    assertThat(response.getContentAsString()).startsWith("[]");
```
jaysaleh, line 123: I came across this as well, maybe we should add a comment as to why we do `startsWith("[]")`. I can see someone who hasn't dealt with the code being confused here

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    datastore.put(createTestCommentEntity(1));
+    datastore.put(createTestCommentEntity(1));
+    datastore.put(createTestCommentEntity(1));
+
+    servlet.doGet(request, response);
+
+    assertThat(response.getContentType()).isEqualTo("application/json;");
+    String json = response.getContentAsString();
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    Type listType = new TypeToken<ArrayList<Comment>>() {}.getType();
```
jaysaleh, line 138: Up to you, this code is used more than once, maybe make a helper function to clean up the test

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    servlet.doPost(request, response);
+
+    assertThat(response.getStatus()).isEqualTo(HttpServletResponse.SC_FORBIDDEN);
+    assertThat(datastore.prepare(new Query(Comment.ENTITY_KIND)).countEntities(withLimit(1)))
+        .isEqualTo(0);
+  }
+
+  @Test
+  public void doPost_storesCommentWithProperties_noParent() throws ServletException, IOException {
+    int LECTURE_ID = 1;
```
amuamushu, line 74: Can you structure your tests to be arrange, act, assert?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    assertThat(comment.lecture().getId()).isEqualTo(LECTURE_ID);
+    assertThat(comment.author().getEmail()).isEqualTo(EMAIL);
+    assertThat(comment.content()).isEqualTo(CONTENT);
+    assertThat(comment.parent().isPresent()).isFalse();
+  }
+
+  @Test
+  public void doPost_storesCommentParent() throws ServletException, IOException {
+    int PARENT_ID = 32;
+    testServices.setEnvIsLoggedIn(true);
```
amuamushu, line 99: arrange, act, assert here too please

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    assertThat(response.getStatus()).isEqualTo(HttpServletResponse.SC_ACCEPTED);
+    Comment comment = Comment.fromEntity(query.asSingleEntity());
+    assertThat(comment.parent().isPresent()).isTrue();
+    assertThat(comment.parent().get().getId()).isEqualTo(PARENT_ID);
+  }
+
+  @Test
+  public void doGet_returnsNothingForUnknownLecture() throws ServletException, IOException {
+    request.setParameter(DiscussionServlet.PARAM_LECTURE, "1");
```
amuamushu, line 118: Should 1 be a constant variable? I feel having the value itself is a bit distracting since I'm not sure what the significance of 1 is. Also, I noticed you used 1 in your other tests so I think it would be nice to define it globally.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    String json = response.getContentAsString();
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    Type listType = new TypeToken<ArrayList<Comment>>() {}.getType();
+    List<Comment> comments = gson.fromJson(json, listType);
+    assertThat(comments.size()).isEqualTo(3);
+  }
+
+  @Test
+  public void doGet_returnsCommentsForSpecificLecture() throws ServletException, IOException {
+    request.setParameter(DiscussionServlet.PARAM_LECTURE, "2");
```
amuamushu, line 145: Can you make 2 a constant? I'm not sure what 2 signifies.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    datastore.put(createTestCommentEntity(1));
+    datastore.put(createTestCommentEntity(1));
```
amuamushu, line 147: I don't think it's clear that `createTestCommentEntity` takes the lecture id as the parameter just from reading the test. Could you either update the method header to use constants instead of numbers?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+    Comment comment = Comment.fromEntity(query.asSingleEntity());
+    assertThat(comment.lecture().getId()).isEqualTo(LECTURE_ID);
+    assertThat(comment.author().getEmail()).isEqualTo(EMAIL);
+    assertThat(comment.content()).isEqualTo(CONTENT);
+    assertThat(comment.parent().isPresent()).isFalse();
+  }
+
+  @Test
+  public void doPost_storesCommentParent() throws ServletException, IOException {
+    int PARENT_ID = 32;
```
amuamushu, line 98: Should this be camel case? It's not a final variable.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+  @After
+  public void cleanUp() {
+    testServices.tearDown();
+  }
+
+  @Test
+  public void doPost_loggedOut_postForbidden() throws ServletException, IOException {
+    testServices.setEnvIsLoggedIn(false);
+    String LECTURE_ID = "1";
```
amuamushu, line 62: Since LECTURE_ID is defined in multiple tests, can you create a constant for them? The 1 is a bit distracting, assuming it doesn't mean anything.

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: IDK.  Gonna wait until we figure out if we're using a util/consts class for all of these

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Should be clearer now

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I just parsed the json into a list and made sure it was empty

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Created helper function

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Done

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Done

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I added some constants, should be clearer now

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I added some constants, should be clearer now

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I added some constants, should be clearer now

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Added it as a constant

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I added some constants, should be clearer now

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: We'll talk about organizing these soon

src/test/java/com/googleinterns/zoomtube/servlets/AuthenticationServletTest.java
```diff
@@ -81,11 +82,7 @@ public void doGet_loggedIn_expectFalse() throws Exception {
 
   @Test
   public void doGet_returnsUserEmail() throws Exception {
-    final String EMAIL = "test@example.com";
-    authService.setEnvIsLoggedIn(true);
-    authService.setEnvAuthDomain("example.com");
-    authService.setEnvEmail(EMAIL);
```
billie-cat, line 87: This is an important part of the test and therefore should be done here. If someone was reading these tests they would need to know that they need to look at setup to find where the email was set.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,202 @@
...
+  private static final int LECTURE_B = 2;
+  private static final int LECTURE_C = 3;
+  private static final String LECTURE_A_STR = "1";
+  private static final String LECTURE_B_STR = "2";
+  private static final String LECTURE_C_STR = "3";
+  private static final int PARENT_ID = 32;
+  private static final String PARENT_ID_STR = "32";
+  private static final String CONTENT = "Test content";
+  private static final String EMAIL = "test@example.com";
+  private static final String AUTH_DOMAIN = "example.com";
```
billie-cat, line 58: I thought we had talked about keeping these in the tests themselves to reduce cognitive load? Why did you move them all up here?

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
billie-cat: So these need to be public because of tests? In that case I would make these private and just duplicate the variables in the tests. The issue with making these public is that you end up allowing other classes to use them. 

+1 to what Jay said. I would try @VisibleForTesting because duplicating the variable in tests.

src/test/java/com/googleinterns/zoomtube/servlets/AuthenticationServletTest.java
n-wach: Duplicated to relevant test

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Talked in chat, made changes.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,204 @@
...
+import org.mockito.junit.MockitoRule;
+
+public class DiscussionServletTest {
+  private static final int LECTURE_ID = 1;
+  private static final String LECTURE_ID_STR = "1";
+  private static final String TEST_COMMENT_CONTENT = "Test content";
+  private static final String DEFAULT_EMAIL = "test@example.com";
+  private static final String AUTH_DOMAIN = "example.com";
+
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
```
f-camp, line 54:
```suggestion
  @Rule private final MockitoRule mockito = MockitoJUnit.rule();
```

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,204 @@
...
+    assertThat(comment.lecture().getId()).isEqualTo(LECTURE_ID);
+    assertThat(comment.author().getEmail()).isEqualTo("author@example.com");
+    assertThat(comment.content()).isEqualTo("Something unique");
+    assertThat(comment.parent().isPresent()).isFalse();
+  }
+
+  @Test
+  public void doPost_storesCommentParent() throws ServletException, IOException {
+    testServices.setEnvIsLoggedIn(true);
+    final int PARENT_ID = 32;
```
f-camp, line 111:
```suggestion
   final int parentId = 32;
```

is fine here.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,204 @@
...
+    verify(response).setContentType("application/json;");
+    List<Comment> comments = getCommentsFromJson(content.toString());
+    assertThat(comments.size()).isEqualTo(3);
+  }
+
+  @Test
+  public void doGet_returnsCommentsForSpecificLecture() throws ServletException, IOException {
+    final int LECTURE_A = 1;
+    final int LECTURE_B = 2;
+    final int LECTURE_C = 3;
```
f-camp, line 162: ditto

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: `org.junit.internal.runners.rules.ValidationError: The @Rule 'mockito' must be public.`

n-wach: No idea why, or if this can be fixed, but I can't run with it private

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Fixed!

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Fixed!

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+public class DiscussionServletTest {
+  private static final int LECTURE_ID = 1;
+  private static final String LECTURE_ID_STR = "1";
+  private static final String TEST_COMMENT_CONTENT = "Test content";
```
billie-cat, line 50: nit: could probably rename this to COMMENT_CONTENT

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  @Test
+  public void doPost_loggedOut_postForbidden() throws ServletException, IOException {
+    testServices.setEnvIsLoggedIn(false);
+    when(request.getParameter(DiscussionServlet.PARAM_LECTURE)).thenReturn(LECTURE_ID_STR);
+
+    servlet.doPost(request, response);
+
+    verify(response).sendError(eq(HttpServletResponse.SC_FORBIDDEN), anyString());
```
billie-cat, line 85: It is better to actually add the string you expect to return. The more specific the test the better.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+    when(request.getParameter(DiscussionServlet.PARAM_LECTURE)).thenReturn(LECTURE_ID_STR);
+    StringWriter content = new StringWriter();
+    PrintWriter writer = new PrintWriter(content);
+    when(response.getWriter()).thenReturn(writer);
+
+    servlet.doGet(request, response);
+
+    verify(response).setContentType("application/json;");
+    List<Comment> comments = getCommentsFromJson(content.toString());
+    assertThat(comments.isEmpty()).isTrue();
```
billie-cat, line 139: you can rewrite it as: assertThat(comments).isEmpty();

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+    datastore.put(createTestCommentEntity(LECTURE_ID));
+    StringWriter content = new StringWriter();
+    PrintWriter writer = new PrintWriter(content);
+    when(response.getWriter()).thenReturn(writer);
+
+    servlet.doGet(request, response);
+
+    verify(response).setContentType("application/json;");
+    List<Comment> comments = getCommentsFromJson(content.toString());
+    assertThat(comments.size()).isEqualTo(3);
```
billie-cat, line 156: you can use: assertThat(comments).hasSize(3);

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  private Entity createTestCommentEntity(int lectureId) {
+    Entity commentEntity = new Entity(Comment.ENTITY_KIND);
+    Key lecture = KeyFactory.createKey(/* kind= */ "Lecture", lectureId);
+    commentEntity.setProperty(Comment.PROP_LECTURE, lecture);
+    commentEntity.setProperty(Comment.PROP_PARENT, null);
+    commentEntity.setProperty(Comment.PROP_TIMESTAMP, new Date(0));
+    commentEntity.setProperty(Comment.PROP_AUTHOR, new User(DEFAULT_EMAIL, AUTH_DOMAIN));
+    commentEntity.setProperty(Comment.PROP_CONTENT, TEST_COMMENT_CONTENT);
```
billie-cat, line 193: Is this data related to the other test which also uses TEST_COMMENT_CONTENT? I would suggest not using the same value as it adds confusion as to if they are related. If they are related then it should be more obvious. Eg, you can pass in this field.

billie-cat: Same thing for DEFAULT_EMAIL and AUTH_DOMAIN

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+    commentEntity.setProperty(Comment.PROP_CREATED, new Date(Clock.systemUTC().millis()));
+
+    return commentEntity;
+  }
+
+  private List<Comment> getCommentsFromJson(String json) {
+    Gson gson = new GsonBuilder().registerTypeAdapterFactory(GenerateTypeAdapter.FACTORY).create();
+    Type listType = new TypeToken<ArrayList<Comment>>() {}.getType();
+    List<Comment> comments = gson.fromJson(json, listType);
```
billie-cat, line 202: can you do: return gson.fromJson(json, listType);

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+    servlet.doPost(request, response);
+
+    verify(response).sendError(eq(HttpServletResponse.SC_FORBIDDEN), anyString());
+    assertThat(datastore.prepare(new Query(Comment.ENTITY_KIND)).countEntities(withLimit(1)))
+        .isEqualTo(0);
+  }
+
+  @Test
+  public void doPost_storesCommentWithProperties_noParent() throws ServletException, IOException {
```
billie-cat, line 91: This test name confuses me. What are you testing? Seems you are testing 2 things, storing comments and not having any parents. Should it be split into 2 tests?

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -47,8 +48,8 @@
 // TODO: References to Lecture kind string literal need to be updated once Lecture code is merged.
 @WebServlet("/discussion")
 public class DiscussionServlet extends HttpServlet {
-  private static final String PARAM_LECTURE = "lecture";
-  private static final String PARAM_PARENT = "parent";
+  @VisibleForTesting protected static final String PARAM_LECTURE = "lecture";
```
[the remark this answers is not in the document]
f-camp: why don't we just make it public instead? Since it is used outside of this class. We should be conservative with @VisibleForTesting and find other options before opting for it.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I check every property except the parent property, and test the parent property in another test.  Maybe I should combine them?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: changed

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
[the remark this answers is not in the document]
n-wach: Changed

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+  private static final String TEST_COMMENT_CONTENT = "Test content";
+  private static final String DEFAULT_EMAIL = "test@example.com";
+  private static final String AUTH_DOMAIN = "example.com";
+
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
+
+  @Mock private HttpServletRequest request;
+  @Mock private HttpServletResponse response;
+  private DiscussionServlet servlet;
+  private LocalServiceTestHelper testServices = new LocalServiceTestHelper(
```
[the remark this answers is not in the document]
n-wach: Changed

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+  @Test
+  public void doPost_loggedOut_postForbidden() throws ServletException, IOException {
+    testServices.setEnvIsLoggedIn(false);
+    when(request.getParameter(DiscussionServlet.PARAM_LECTURE)).thenReturn(LECTURE_ID_STR);
+
+    servlet.doPost(request, response);
+
+    verify(response).sendError(eq(HttpServletResponse.SC_FORBIDDEN), anyString());
+    assertThat(datastore.prepare(new Query(Comment.ENTITY_KIND)).countEntities(withLimit(1)))
+        .isEqualTo(0);
```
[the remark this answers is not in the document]
n-wach: We can only get the count, an iteratable, or a single entity (which can be null if there is none).  I think this is most clear.  Also, isEmpty doesn't seem to be defined in my version of google truth?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Not supported by this version of truth

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Not supported by this version of truth

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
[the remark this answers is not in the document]
n-wach: Not supported by this version of truth

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I moved things around so it's more clear what I'm testing

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: I removed these constants as there were only used in one place and their values were never tested anywhere,

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Removed it

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Done

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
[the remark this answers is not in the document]
billie-cat: @f-camp isn't making it public because of tests less ideal than @VisibleForTesting?

src/test/java/com/googleinterns/zoomtube/servlets/AuthenticationServletTest.java
```diff
@@ -29,26 +29,28 @@
   private AuthenticationServlet servlet;
   @Mock private HttpServletRequest request;
   @Mock private HttpServletResponse response;
```
f-camp, line 31: nit: It is good practice to have mocks at the beginning of tests.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,204 @@
...
+public class DiscussionServletTest {
+  private static final int LECTURE_ID = 1;
+  private static final String LECTURE_ID_STR = "1";
+  private static final LocalServiceTestHelper testServices = new LocalServiceTestHelper(
+      new LocalUserServiceTestConfig(), new LocalDatastoreServiceTestConfig());
+
+  @Rule public final MockitoRule mockito = MockitoJUnit.rule();
+
+  @Mock private HttpServletRequest request;
+  @Mock private HttpServletResponse response;
```
f-camp, line 56: nit: it is good practice to have @Rule and @Mocks at the beginning of tests.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -0,0 +1,204 @@
...
+  private DatastoreService datastore;
+
+  @Before
+  public void setUp() throws ServletException {
+    testServices.setUp();
+    testServices.setEnvEmail("test@example.com");
+    testServices.setEnvAuthDomain("example.com");
+    servlet = new DiscussionServlet();
+    servlet.init();
+    datastore = DatastoreServiceFactory.getDatastoreService();
```
f-camp, line 67: Is it the same datastore here that is used form within the servlet init as well?

n-wach: Because I created a new LocalServiceTestHelper, any call to .getDatastoreService will return the test service.  And because I .setUp before calling servlet.init() (which does .getDatastoreService()), both the servlet and the tests will be using the same test datastore.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Moved

src/test/java/com/googleinterns/zoomtube/servlets/AuthenticationServletTest.java
n-wach: Moved

PR #89, opened 2020-07-31
Update hard-coded test JSON to use Date instead of long.

amuamushu:
After switching to storing the start, end, and duration as dates, the hard-coded JSON used for testing is no longer accurate. This pull request updates the hard-coded strings and helper methods to reflect this change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,14 +151,14 @@
         <configuration>
           <annotationProcessorPaths>
             <path>
-              <groupId>com.ryanharter.auto.value</groupId>
-              <artifactId>auto-value-gson-extension</artifactId>
-              <version>1.3.0</version>
-            </path>
-            <path>
               <groupId>com.google.auto.value</groupId>
               <artifactId>auto-value</artifactId>
               <version>1.7.3</version>
+            </path>
+            <path>
+              <groupId>com.ryanharter.auto.value</groupId>
+              <artifactId>auto-value-gson-extension</artifactId>
+              <version>1.3.0</version>
             </path>
           </annotationProcessorPaths>
         </configuration>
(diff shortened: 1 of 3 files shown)


src/main/webapp/transcript.js
```diff
@@ -67,9 +67,9 @@ function appendTextToList(transcriptLine, ulElement) {
   const startDate = new Date(transcriptLine.start);
   const endDate = new Date(transcriptLine.end);
   const startTimestamp = `${startDate.getHours()}:${startDate.getMinutes()}:${
-      startDate.getSeconds()}`;
+    startDate.getSeconds()}`;
   const endTimestamp =
-      `${endDate.getHours()}:${endDate.getMinutes()}:${endDate.getSeconds()}`;
+    `${endDate.getHours()}:${endDate.getMinutes()}:${endDate.getSeconds()}`;
```
n-wach, line 72: Maybe these could be a util function? dateToTimestampString

pom.xml
```diff
@@ -150,16 +150,16 @@
...
             <path>
               <groupId>com.google.auto.value</groupId>
               <artifactId>auto-value</artifactId>
               <version>1.7.3</version>
             </path>
+            <path>
+              <groupId>com.ryanharter.auto.value</groupId>
+              <artifactId>auto-value-gson-extension</artifactId>
+              <version>1.3.0</version>
+            </path>
```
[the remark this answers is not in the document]
amuamushu: Thanks! fixed

src/main/webapp/transcript.js
amuamushu: I'm creating a Util class in a different pull request right now. I added an issue for this #94 so I won't forget. Thanks!

n-wach: Sounds good!

PR #111, opened 2020-08-05
Arrange the layout of view.html using bootstrap

n-wach:
This arranges the basic layout of the lecture view page. 
Here is a screenshot: https://screenshot.googleplex.com/ijxAPNhm56f

This doesn't address detailed styling of the various components.  Styling on discussion will follow soon.


src/main/webapp/style.css
```diff
@@ -107,9 +121,33 @@ html {
...
+
+.halfpage {
+  height: calc(100vh - 56px);
+  max-height: calc(100vh - 56px);
+  max-width: 50vw;
+  width: 50vw;
+  overflow: hidden;
+}
+
+#page {
```
amuamushu, line 134: Not too sure if this would be here but could you add padding to the top of the page div? The video and comment section seem very close to the nav bar.

n-wach: I'm going to hold off on that for a future PR.  This was mostly the layout of the left/right half-pages.

n-wach: Added to description

src/main/webapp/style.css
```diff
@@ -1,38 +1,39 @@
...
+  src: url('./fonts/product-sans/ProductSansRegular.ttf') format('truetype');
 }
 
-body {
-  margin-bottom: 60px !important;
+* {
+  font-family: 'Product Sans';
 }
 
+body,
```
billie-cat, line 14: I believe these should be alphabetized. 

https://google.github.io/styleguide/htmlcssguide.html#Declaration_Order

src/main/webapp/style.css
```diff
@@ -88,8 +86,24 @@ html {
...
+#submit-btn-cont .btn {
+  background-color: #4c8bf5 !important;
+}
+
+#add-btn-cont .btn,
+#add-btn-cont .btn:hover {
+  width: 200px !important;
+}
+
+#add-btn-cont .btn:hover,
```
billie-cat, line 104: Is it supposed to be formatted this way? Can you look into this from the style guide?

src/main/webapp/style.css
n-wach: I think that only applies to the declarations within the brackets--the order of selectors matters somewhat when overriding prior ones.  But this is an easy change and consistent with everything else so I'll switch em :)

src/main/webapp/style.css
n-wach: Yeah the `selector` continues until the comma, which should all be on one line

src/main/webapp/style.css
```diff
@@ -1,38 +1,39 @@
...
+/* Global styling */
 
-body {
-  margin-bottom: 60px !important;
-}
-
-html {
-  min-height: 100%;
-  position: relative;
+@font-face {
```
f-camp, line 3: For this whole file I would look into alphabetizing all the declarations. 
Google code is usually arranged in alphabetical order for every kind of declaration and this is done recursively.
https://google.github.io/styleguide/htmlcssguide.html#Declaration_Order

src/main/webapp/style.css
billie-cat: Can you add Sahil to this review. I feel he would be better for this stuff.

src/main/webapp/style.css
n-wach: We will organize/split CSS in a future PR, if that's ok

billie-cat: Why not split it now?

n-wach: There are a lot of other PRs out rn that touch this file and it would complicate a bunch of merges.  I could create a new PR with everything split, but we'd have to redo all of the styling merges again...  I was gonna make the whole directory restructure it's own PR

n-wach: Tracking issue #123

PR #112, opened 2020-08-05
Add seeking functions for syncing

jaysaleh:
This pr sets up the framework for synchronization.


src/main/webapp/view.js
```diff
@@ -15,22 +15,54 @@
...
 }
 
+/**
+ * Retrives the current time of the video and relays time to
+ * {@code sync}.
+ */
+function getCurrentTime() {
+  const currentTime = window.videoPlayer.getCurrentTime();
+  // If already synced to this time, don't sync again.
+  if (lastSyncedTime == currentTime) {
```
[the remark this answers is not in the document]
jaysaleh: Isn't this way more efficient

src/main/webapp/view.js
```diff
@@ -15,22 +15,54 @@
 const PARAM_ID = 'id';
 const PARAM_VIDEO_ID = 'video-id';
 
+let timer;
```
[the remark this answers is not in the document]
jaysaleh: removed

src/main/webapp/view.js
```diff
@@ -15,22 +15,54 @@
...
+  // TODO: Adjust time interval.
+  timer = window.setInterval(getCurrentTime, 1000);
   // TODO: Initialize the trancript section.
 }
 
+/**
+ * Retrives the current time of the video and relays time to
+ * {@code sync}.
+ */
+function getCurrentTime() {
```
amuamushu: can this be updated to `getCurrentVideoTime`? I think it can get a bit confusing later since the comments and transcripts also have timestamps. Also, can you specify the time units for this in the method header?

src/main/webapp/view.js
```diff
@@ -15,22 +15,54 @@
...
+  if (lastSyncedTime == currentTime) {
+    return;
+  }
+  lastSyncedTime = currentTime;
+  sync(currentTime, false);
+}
+
+/**
+ * Calls functions that seek video, transcript,
+ * and discussion to {@code currentTime}.
```
amuamushu, line 50: Can you mention `SyncVideo` in the method header please?

src/main/webapp/view.js
```diff
@@ -15,22 +15,53 @@
...
+  // TODO: Adjust time interval.
+  window.timer = window.setInterval(syncEverythingToVideo, 1000);
   // TODO: Initialize the trancript section.
 }
 
+/**
+ * Retrives the current time of the video and relays time to
+ * {@code sync}.
+ */
+function syncEverythingToVideo() {
```
amuamushu, line 43: Do you think you can combine `syncEverythingToVideo` with `sync`? I find it redundant that there's two methods for syncing. Since there's already a boolean for the video in `sync`, I think you can have a check at the beginning to see if its the video and update the current time there.

src/main/webapp/view.js
```diff
@@ -15,22 +15,53 @@
...
+  const currentTime = window.videoPlayer.getCurrentTime();
+  sync(currentTime, false);
+}
+
+/**
+ * Calls functions that seek video, transcript,
+ * and discussion to {@code currentTime}.
+ */
+function sync(currentTime, syncVideo = false) {
```
n-wach, line 52: Oops, i'm sorry.  I think this should actually default to true, so discussion and transcript don't need to specify it...

src/main/webapp/view.js
```diff
@@ -15,22 +15,53 @@
...
- * Initializes the video player and discussion section
+ * Initializes the timer, video player, and discussion section
  * for the lecture view page.
  */
 async function initialize() {
   window.loadVideoApi();
-  window.intializeDiscussion();
+  window.loadDiscussion();
+  // TODO: Adjust time interval.
+  window.timer = window.setInterval(syncEverythingToVideo, 1000);
```
n-wach, line 35: Since this function does so little, what do you think about inlining it?
```suggestion
  window.timer = window.setInterval(() => {
    sync(window.videoPlayer.getCurrentTime(), false);
  }, 1000);
```

src/main/webapp/view.js
n-wach: I agree, another option to consider is in-lining it in the `setInterval`

jaysaleh: made inline

src/main/webapp/view.js
```diff
@@ -15,22 +15,47 @@
...
+  window.timer = window.setInterval(() => {
+    sync(window.videoPlayer.getCurrentTime(), false);
+  }, /* ms */ 1000);
   // TODO: Initialize the trancript section.
 }
 
+
+/**
+ * Calls functions that seek video (if {@code syncVideo} is true),
+ * transcript, and discussion to {@code currentTime}.
```
n-wach, line 44: Could you specify what the units of `currentTime` are?  A number representing the total number of seconds since the start of the video?

src/main/webapp/view.html
```diff
@@ -11,6 +11,7 @@
   crossorigin="anonymous"
 />
 <script src="/discussion.js" defer></script>
+<script src="/transcript.js" defer></script>
```
amuamushu, line 14: There is already a `transcript.js` in this html file. Can you remove one of them?

src/main/java/com/googleinterns/zoomtube/servlets/DeleteTranscriptServlet.java
```diff
@@ -31,7 +31,7 @@
 @WebServlet("/delete-transcript")
 public class DeleteTranscriptServlet extends HttpServlet {
   private static String HOME_PATH = "/";
-  
+
```
billie-cat, line 34: this should have been done in its own PR as it is unrelated.

src/main/webapp/transcript.js
```diff
@@ -66,7 +66,7 @@ function appendTextToList(transcriptLine, ulElement) {
   const startTimestamp = `${startDate.getHours()}:${startDate.getMinutes()}:${
     startDate.getSeconds()}`;
   const endTimestamp =
-    `${endDate.getHours()}:${endDate.getMinutes()}:${endDate.getSeconds()}`;
+      `${endDate.getHours()}:${endDate.getMinutes()}:${endDate.getSeconds()}`;
```
billie-cat, line 69: is this change a mistake?

src/main/webapp/video.js
```diff
@@ -45,3 +43,31 @@ function onYouTubeIframeAPIReady() {
 function onPlayerReady(event) {
   event.target.playVideo();
 }
+
+/** Starts timer if {@code currState} is playing, stops otherwise. */
```
billie-cat, line 47: please rename currState to currentState.

It is recommended to not use abbreviations, see https://docs.google.com/document/d/1_ldClEZLsI1cMpQl_BfeIlkjTsE2IiJ2PgvhABfO6vs/edit#heading=h.wx9q7rdncnvo

src/main/webapp/video.js
```diff
@@ -45,3 +43,31 @@ function onYouTubeIframeAPIReady() {
...
+function startStopTimerOnStateChange(currState) {
```
billie-cat, line 48: for this method you should just pass state as that is all you care about.

src/main/webapp/video.js
```diff
@@ -45,3 +43,31 @@ function onYouTubeIframeAPIReady() {
...
+  const state = currState.data;
+  if (state == window.YT.PlayerState.PLAYING) {
+    startTimer();
+    return;
+  }
+  stopTimer();
+}
+
+/** Starts timer which calls {@code sync} with the current video time.  */
```
billie-cat, line 57: This comment describes the code, not the functionality. For me to understand what this actually does I would need to look up the JSDoc for sync.

src/main/webapp/video.js
```diff
@@ -45,3 +43,31 @@ function onYouTubeIframeAPIReady() {
...
+function startTimer() {
+  window.timer = window.setInterval(() => {
+    window.sync(window.videoPlayer.getCurrentTime(), false);
+  }, /* ms */ 1000);
```
billie-cat, line 61:
```suggestion
  }, /* ms= */ 1000);
```

src/main/webapp/video.js
```diff
@@ -45,3 +43,31 @@ function onYouTubeIframeAPIReady() {
...
+    startTimer();
+    return;
+  }
+  stopTimer();
+}
+
+/** Starts timer which calls {@code sync} with the current video time.  */
+function startTimer() {
+  window.timer = window.setInterval(() => {
+    window.sync(window.videoPlayer.getCurrentTime(), false);
```
billie-cat, line 60: Also please add the variable comment for false. I do not know what this means.

src/main/java/com/googleinterns/zoomtube/servlets/DeleteTranscriptServlet.java
jaysaleh: The linter did this

src/main/webapp/transcript.js
jaysaleh: I needed to make this change for the js validation to work

src/main/webapp/video.js
jaysaleh: fixed

src/main/webapp/video.js
jaysaleh: The api determines what to pass to this function. In this case they send an object with `.data` having the video state

src/main/webapp/video.js
```diff
@@ -45,3 +43,30 @@ function onYouTubeIframeAPIReady() {
...
+  if (currentState.data == window.YT.PlayerState.PLAYING) {
+    startTimer();
+    return;
+  }
+  stopTimer();
+}
+
+/** Starts timer which calls {@code sync} with the current video time.  */
+function startTimer() {
+  window.timer = window.setInterval(() => {
```
n-wach, line 58: can we call this `videoSyncTimer` or something a little less general

n-wach: also consider renaming the start/stop functions.  unfortunately they are added to the global/window scope, and i worry there could be name collisions with other timers in the future

src/main/webapp/video.js
jaysaleh: fixed

src/main/java/com/googleinterns/zoomtube/servlets/DeleteTranscriptServlet.java
billie-cat: It is weird it didn't fix this before it was merged.

src/main/webapp/transcript.js
billie-cat: what do you mean by validation? the linter?

src/main/webapp/video.js
```diff
@@ -45,3 +43,18 @@ function onYouTubeIframeAPIReady() {
 function onPlayerReady(event) {
   event.target.playVideo();
 }
+
+/** Starts timer if {@code currentState} is playing, stops otherwise. */
+function startStopTimerOnStateChange(currentState) {
```
billie-cat, line 48: nit: I would name this method startOrStopTimer(). You can omit the state change because you can deduce that from the input.

src/main/webapp/video.js
billie-cat: sgtm

src/main/webapp/transcript.js
jaysaleh: Yes

PR #114, opened 2020-08-05
Use Bootstrap to make discussion section match design mock

n-wach:
This uses bootstrap and templates to make the Discussion section match what we had planned in our design mock.

Screenshot:  https://screenshot.googleplex.com/n9CYmDtTo97

Once transcript DOM is added, we can talk about making styling better/consistent, so hold off on minor styling comments for now


src/main/webapp/discussion.js
```diff
@@ -103,53 +129,54 @@ async function fetchDiscussion() {
...
-  const repliesDiv = document.createElement('div');
+class DiscussionComment extends HTMLElement {
+  constructor(comment) {
+    super();
+    this.attachShadow({mode: 'open'});
+    const shadow = TEMPLATE_COMMENT.content.cloneNode(true);
+    this.shadowRoot.appendChild(shadow);
+
+    const username = comment.author.email.split('@')[0];
+    const timestampPrefix = comment.parentKey.value ? '' : '00:00 - ';
```
amuamushu, line 142: Can you add a TODO to update `00:00` with the timestamp once it's available?

src/main/webapp/discussion.js
```diff
@@ -103,53 +129,54 @@ async function fetchDiscussion() {
...
+    const header = `${timestampPrefix}${username} on ${comment.created}`;
+    this.setSlotSpan(SLOT_HEADER, header);
+
+    this.setSlotSpan(SLOT_CONTENT, comment.content);
+
+    const replies = document.createElement('div');
```
amuamushu, line 148: nit: would it be beneficial to create helper methods for the constructor? I feel like the code for creating replies can be it's own helper method.

src/main/webapp/discussion.js
n-wach: I moved this into its own function with a TODO, thanks!

src/main/webapp/discussion.js
n-wach: Done

src/main/webapp/style.css
```diff
@@ -50,6 +46,20 @@ body {
   background-color: #f3f3f3;
 }
 
+.btn-wide {
```
jaysaleh, line 49: since these share code you could put them together and differentiate the width

n-wach: Done

src/main/webapp/style.css
```diff
@@ -36,18 +36,22 @@ body {
 .btn {
```
f-camp, line 36: Alphabetize this file please.
Should be done for all declarations recursively.
https://google.github.io/styleguide/htmlcssguide.html#Declaration_Order

src/main/webapp/discussion.js
```diff
@@ -22,13 +22,39 @@ const ATTR_ID = 'key-id';
...
 
+/**
+ * Posts a new comment using the main post textarea.
+ */
+async function postNewComment() {
+  postAndReload(ELEMENT_POST_TEXTAREA);
+}
+
+/**
+ * Posts a reply to a comment.
```
billie-cat, line 52: Please include information about what textarea and parentId represent.

src/main/webapp/discussion.js
```diff
@@ -22,13 +22,39 @@ const ATTR_ID = 'key-id';
...
+ */
+async function postReply(textarea, parentId) {
```
billie-cat, line 54: seems textarea should be textArea

src/main/webapp/discussion.js
```diff
@@ -103,53 +129,77 @@ async function fetchDiscussion() {
...
+    this.shadowRoot.appendChild(shadow);
+
+    const username = comment.author.email.split('@')[0];
+    let timestampPrefix = '';
+    if (!comment.parentKey.value) {
+      // Only display timestamp on root comments.
+      timestampPrefix = `${this.timestampToString(comment.timestamp)} - `;
+    }
+    const header = `${timestampPrefix}${username} on ${comment.created}`;
+    this.setSlotSpan(SLOT_HEADER, header);
```
billie-cat, line 148: typically, constructors should never do any work. they should only build the object.

src/main/webapp/discussion.js
```diff
@@ -103,53 +129,77 @@ async function fetchDiscussion() {
...
+  // TODO: Implement using a utility function.
+  timestampToString(timestamp) {
+    const ts = new Date(timestamp);
+    const seconds = ts.getSeconds().toString().padStart(2, '0');
+    const minutes = ts.getMinutes().toString().padStart(2, '0');
+    return `${minutes}:${seconds}`;
+  }
+
+  /**
+   * Creates a Discussion Comment for every reply to this comment, and adds
```
billie-cat, line 180:
```suggestion
   * Creates a {@code DiscussionComment} for {@code replies} to this comment, and adds
```

billie-cat: I am a little confused by your doc. When you wrote: "to this comment"

What comment are you referring to?

src/main/webapp/discussion.js
```diff
@@ -103,53 +129,77 @@ async function fetchDiscussion() {
...
+   * them to a {@code div} in the replies slot.
```
billie-cat, line 181: What is "replies slot"

src/main/webapp/discussion.js
```diff
@@ -103,53 +129,77 @@ async function fetchDiscussion() {
...
-  const parentId = repliesDiv.parentElement.getAttribute(ATTR_ID);
-  submit.innerText = 'Post';
-  submit.onclick = () => {
-    postAndReload(textarea, parentId);
-  };
-  div.appendChild(textarea);
-  div.appendChild(submit);
-
-  repliesDiv.prepend(div);
+  /**
```
billie-cat, line 168: Should this method move to that Utils class you were talking about?

src/main/webapp/discussion.js
billie-cat: You could move this logic to a method called "createFirstComment" or "initializeCommentForm" or some other method name depending on what you are doing.

billie-cat: This could probably even be split into multiple methods. Seems like a lot is happening here.

src/main/webapp/discussion.js
n-wach: Will do once utils are merged--See TODO

src/main/webapp/discussion.js
n-wach: The `DiscussionComment` represents a comment, and there may be replies to that comment.  so the comment i'm referring to is the one the class instance represents (the one passed via the constructor).  Will clarify

src/main/webapp/discussion.js
n-wach: Will do

src/main/webapp/discussion.js
n-wach: I think that would be confusing, as the element tag is always written as `textarea` : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

n-wach: I think i'll change it to inputField, as it doesn't need to be a textarea

src/main/webapp/style.css
n-wach: We will organize/split CSS in a future PR, if that's ok

src/main/webapp/discussion.js
n-wach: Done

src/main/webapp/discussion.js
n-wach: Done

src/main/webapp/discussion.js
n-wach: Templates have a number of named slots that can be filled with other elements.  I will change some things around to make it more clear

src/main/webapp/discussion.js
```diff
@@ -103,53 +129,77 @@ async function fetchDiscussion() {
...
+    const minutes = ts.getMinutes().toString().padStart(2, '0');
+    return `${minutes}:${seconds}`;
+  }
+
+  /**
+   * Creates a Discussion Comment for every reply to this comment, and adds
+   * them to a {@code div} in the replies slot.
+   */
+  createReplies(replies) {
+    const replyDiv = document.createElement('div');
```
billie-cat, line 184: Where is document defined?

src/main/webapp/discussion.js
```diff
@@ -103,53 +129,77 @@ async function fetchDiscussion() {
 }
 
 /**
- * Creates an element for displaying {@code comment}.
+ * Renders a comment and its replies, with a form to post a new reply.
  */
-function createComment(comment) {
-  const element = document.createElement('li');
-  element.setAttribute(ATTR_ID, comment.commentKey.id);
+class DiscussionComment extends HTMLElement {
```
billie-cat, line 134: This class confuses me. It changes all these variables, but hold no internal state.

src/main/webapp/discussion.js
n-wach: document is a javascript global representing the current HTML document--it provides utils for manipulating the DOM

src/main/webapp/discussion.js
n-wach: So this is a custom element template class, meaning it has a number of `slots` defined in the template.  It adds elements to itself, each specifying a `slot` to fill.  I attach the template as shadow-DOM which means this displays as the template, filling in all of the slots.  This class is purely for converting a `comment` into an element displayed on the page

src/main/webapp/style.css
```diff
@@ -131,23 +144,34 @@ body {
...
+#new-comment-form,
+#reply-form {
   width: 100%;
 }
 
+#reply-form {
+  margin-top: 20px;
+}
+
 #discussion-comments {
```
f-camp, line 172: Let's order discussion-comments properly. Even if we want to have "#" together. (Spoke to Sahil a bit about this file).

src/main/webapp/style.css
billie-cat: That seems fine. Can you create a tracking bug?

n-wach: Tracking issue #123

src/main/webapp/style.css
n-wach: Done!

src/main/webapp/discussion.js
n-wach: Done

src/main/webapp/view/discussion/discussion.js
```diff
@@ -103,57 +129,89 @@ async function fetchDiscussion() {
...
  */
-function createComment(comment) {
-  const element = document.createElement('li');
-  element.setAttribute(ATTR_ID, comment.commentKey.id);
+class DiscussionComment extends HTMLElement {
+  /**
+   * @param comment The comment from the servlet that this element should
+   *     render.
+   */
+  constructor(comment) {
```
billie-cat, line 139: This constructor is doing a lot. Can you add a high level comment explaining what it is doing? Not just the @param comment?

n-wach: Done!

PR #117, opened 2020-08-05
Add timestamps.js for timestamp-related functions used in multiple other javascript files

n-wach:
We can add more functions in the future, for now the only ones are:

- `timestampToString` for formatting a timestamp as MM:SS
- `timestampToSeconds` for converting a timestamp to a number of seconds since the start of video


src/main/webapp/util.js
```diff
@@ -0,0 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Coverts a Date representing a video {@code timestamp} into a string.
+ */
+function timestampToString(timestamp) {
```
amuamushu, line 18: I forgot to mention this earlier but I think this should include hours to account for longer lectures.

n-wach: Added, thanks!

src/main/webapp/timestamps.js
```diff
@@ -0,0 +1,39 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Coverts a Date representing a video {@code timestamp} into a string.
+ */
+function timestampToString(timestamp) {
+  const ts = new Date(timestamp);
```
billie-cat, line 19: Please avoid abbreviations, I have mentioned the naming best practices. Please look at those: https://docs.google.com/document/d/1_ldClEZLsI1cMpQl_BfeIlkjTsE2IiJ2PgvhABfO6vs/edit#

Seems a better name is date because that is what it is.

src/main/webapp/timestamps.js
```diff
@@ -0,0 +1,39 @@
...
+  const hours = ts.getHours().toString();
+  return `${hours}:${minutes}:${seconds}`;
+}
+
+/**
+ * Returns the total number of seconds since the start of a video at a
+ * certain {@code timestamp}.
+ */
+function timestampToSeconds(timestamp) {
+  const ts = new Date(timestamp);
```
billie-cat, line 35: here too, use date

src/main/webapp/timestamps.js
n-wach: Fixed, thanks

src/main/webapp/timestamps.js
n-wach: Fixed

PR #120, opened 2020-08-06
Update the style of the transcript to match the design doc.

amuamushu:
The transcript is now updated to reflect the transcript layout on the design document. In this update, users are able to manually scroll. Automatic scrolling with the video will be implemented later.
https://screenshot.googleplex.com/R81F3w57V1j.png


src/main/webapp/transcript.js
```diff
@@ -56,27 +57,28 @@ function addMultipleTranscriptLinesToDom(transcriptLines) {
...
 
-  liElement.startDate = startDate;
-  liElement.endDate = endDate;
+  const liElement = document.createElement('li');
+  liElement.classList.add('align-self-center', 'mb-2');
+  liElement.appendChild(contentDivElement);
+  const hrElement = document.createElement('hr');
+  hrElement.classList.add('my-1', 'align-middle', 'mr-5');
+  liElement.appendChild(hrElement);
+  ulElement.appendChild(liElement);
```
n-wach, line 81: This section is a little hard to follow.  If the transcript styling gets more complicated (like when we add references/icons for comments) we might want to consider making this a customElement using a template

jaysaleh:
Please resolve liner errors before merging

src/main/webapp/transcript.js
amuamushu: I created an issue (#122) for this and will get to it later. Thanks for the tip!

amuamushu:
> Please resolve linter errors before merging

Thanks for noticing. The linter issue is from patching something from a different branch. It should be resolved once it is merged into master.

src/main/webapp/transcript.js
```diff
@@ -102,3 +104,14 @@ function appendParagraphToContainer(text, container, classes = []) {
 function deleteTranscript() {
   fetch('/delete-transcript', {method: 'POST'});
 }
+
+/**
+ * Coverts a Date representing a video {@code timestamp} into a string.
+ */
+// TODO: Move function to timestamp.js
+function timestampToString(timestamp) {
+  const ts = new Date(timestamp);
```
billie-cat, line 113: Please do not use abbreviations. Also this is a date, not a timestamp, so a better name would be date.

"Use the shortest possible name, but not at the cost of clarity. Avoid acronyms, abbreviations, and in-team jargon."

Please look at: https://docs.google.com/document/d/1_ldClEZLsI1cMpQl_BfeIlkjTsE2IiJ2PgvhABfO6vs/edit

src/main/webapp/transcript.js
```diff
@@ -102,3 +104,14 @@ function appendParagraphToContainer(text, container, classes = []) {
...
+  const seconds = ts.getSeconds().toString().padStart(2, '0');
```
billie-cat, line 114: for 2 and '0' can you add the comments for your argument literals

https://docs.google.com/document/d/1-mH8Tg22ioPqe2x6_fjbUml8KkpbO0-KuSDLaql2VK4/edit#

src/main/webapp/style.css
```diff
@@ -107,6 +107,18 @@ html {
   width: 50%;
 }
 
+/* Transcript */
```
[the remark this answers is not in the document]
amuamushu: Yes, I was waiting for #133 to be merged into master to make that change since the `lecture-view.css` file doesn't exist in master yet. I added file and moved the code over for now. 

Done! 

Edit: It was just added so I resolve the merge conflict.

src/main/webapp/transcript.js
amuamushu: Thanks for the info!
This was a patch from an different pull request that I added in to ensure this pull request compiles. 

I believe it is now fixed and added to timestamp.js instead.

src/main/webapp/transcript.js
amuamushu: This wasn't fixed in the other pull request I patched from so I just made the change. Thanks!

src/main/webapp/transcript.js
billie-cat: I see, I would recommend not doing. If you absolutely must please add a review comment indicating this is not your code or else I end up reviewing the code twice.

amuamushu: Will do, sorry about that!

PR #128, opened 2020-08-06
Use unescapeXml when parsing XML for transcript

n-wach:
This means escaped xml characters (like `&#39;` for `'`) are correctly stored and rendered.

While fixing this, another issue was found that was contributing to this problem.  `setContentType` is used to preprocess the content.  We had to remove the semicolon from all content types for them to be correctly identified

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,9 +75,9 @@
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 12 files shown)


n-wach:
Fixed due to another issue--updated PR description

PR #150, opened 2020-08-10
Modifies LectureServlet to respond with one lecture 

jaysaleh:
This pr modifies the `LectureServlet` to respond with one lecture.


src/main/webapp/view/view.js
```diff
@@ -31,6 +34,14 @@ async function initialize() {
   window.loadTranscript(window.location.search);
 }
 
+/** Returns lecture in database associated with {@code window.LECTURE_ID}. */
+async function getLecture() {
+  const url = new URL('/lecture', window.location.origin);
```
n-wach, line 39: `ENDPOINT_LECTURE` const + add to comment

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -99,6 +106,37 @@ public void doPost_urlNotInDatabase_shouldAddToDatabaseAndReturnRedirect() throw
     verify(response).sendRedirect("/view/?id=1&video-id=wXhTHyIgQ_U");
   }
 
+  @Test
+  public void doGet_lectureInDatabase_shouldWriteLecture() throws IOException {
+    Entity lectureEntity = LectureUtil.createEntity(/* lectureName= */ "", TEST_LINK, TEST_ID);
```
n-wach, line 111: Can you test with a non-empty string for name?

src/main/webapp/view/view.js
```diff
@@ -19,7 +19,10 @@ const PARAM_VIDEO_ID = 'video-id';
 window.LECTURE_ID = getLectureId();
 window.VIDEO_ID = getVideoId();
 
-initialize();
+getLecture().then((lecture) => {
```
amuamushu, line 22: Do you think you can switch to only using await or only using .then()? I think it's bad practice to mix them
.
https://docs.google.com/document/d/1J1s2uxmLxeuS6dEj110weXz1ajhAJRhUk0C7fowrNrc/edit

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -67,6 +67,7 @@
   private PrintWriter writer;
 
   private static final String LINK_INPUT = "link-input";
```
amuamushu, line 69: should link-input be camel case?

jaysaleh: link-input is a the input name in the HTML

The style guide didn't include conventions for naming inputs so I defaulted to the convention for ID and class naming.
https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters

src/main/webapp/view/view.js
```diff
@@ -14,12 +14,17 @@
...
 const PARAM_ID = 'id';
 const PARAM_VIDEO_ID = 'video-id';
+const ENDPOINT_LECTURE = '/lecture';
 
 /* exported LECTURE_ID */
 window.LECTURE_ID = getLectureId();
 window.VIDEO_ID = getVideoId();
 
-initialize();
+// TODO: Ask if calling .then() on async is bad practice.
```
f-camp, line 23: TODOs should be actionable code changes, this seems like something that needs to be done before the review.

f-camp:
This doesn't look like we're reviewing for merging to the ```master``` branch. We should be doing reviews for the ```master``` branch merge only otherwise it seems we might need to re-review this when its time to merge.

src/main/webapp/view/view.js
jaysaleh: go/js-practices/async-functions top of page three does something similar

PR #153, opened 2020-08-10
Update transcript methods to use the Lecture Key rather than the Lecture Id.

amuamushu:
This change is needed because `lectureServlet` keeps a reference to the lecture key rather than the lecture id. When `lectureServlet` calls the method from `TranscriptServlet` for parsing the transcript, it would only be able to pass the key.

### DIFF
--- a/src/main/java/com/googleinterns/zoomtube/servlets/TranscriptServlet.java
+++ b/src/main/java/com/googleinterns/zoomtube/servlets/TranscriptServlet.java
@@ -69,7 +69,8 @@ public class TranscriptServlet extends HttpServlet {
     String videoId = request.getParameter(LectureUtil.VIDEO_ID);
     Document document = getTranscriptXmlAsDocument(videoId).get();
     long lectureId = Long.parseLong(request.getParameter(LectureUtil.ID));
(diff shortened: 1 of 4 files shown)


src/main/java/com/googleinterns/zoomtube/servlets/TranscriptServlet.java
```diff
@@ -69,7 +69,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String videoId = request.getParameter(LectureUtil.VIDEO_ID);
     Document document = getTranscriptXmlAsDocument(videoId).get();
     long lectureId = Long.parseLong(request.getParameter(LectureUtil.ID));
-    putTranscriptLinesInDatastore(lectureId, document);
+    Key lectureKey = KeyFactory.createKey(LectureUtil.KIND, lectureId);
```
f-camp, line 72: nit: I'm wondering if  ```LectureUtil.KIND``` needs to be named ```LectureUtil.KEY```?

amuamushu: I feel `KIND` would work better because `KIND` is also used to indicate the type of entity.

PR #155, opened 2020-08-10
Automatically Scroll the Transcript as a Video Plays Forward Without Seeking Backwards

amuamushu:
The transcript now automatically scrolls as the video is playing.

Video of auto scrolling: https://drive.google.com/file/d/1SQaRidffvOQgaoPJA58chcxL5Mrwok1T/view?usp=sharing

Issues that will be fixed later:
- Transcript sometimes scrolls to the next line prematurely
  - This can be mitigated by having the current transcript line be the second line visible in the transcript box rather than the first.
- Users can not scroll on their own while the video is playing. The transcript will scroll back to where it is supposed to be once it is time to scroll to the next line.

### DIFF
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -119,7 +119,8 @@ function deleteTranscript() {
 
 /** Seeks transcript to {@code currentTime}, which is given in seconds. */
 function seekTranscript(currentTime) {
-  const currentTimestamp = window.getDateInSeconds(currentTranscriptLine.endDate);
+  const currentTimestamp =
+      window.getDateInSeconds(currentTranscriptLine.endDate);
   if (currentTime <= currentTimestamp) {
     return;
   }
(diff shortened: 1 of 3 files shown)


src/main/webapp/view/transcript/transcript.js
```diff
@@ -85,6 +86,11 @@ function appendTextToList(transcriptLine, ulElement) {
   ulElement.appendChild(liElement);
   // Save the dates for seeking later.
   liElement.startDate = new Date(transcriptLine.start);
+  liElement.endDate = new Date(transcriptLine.end);
```
jaysaleh, line 89: Why do we need a date here?

amuamushu: The start date is stored so that when the Transcript calls `seekAll`, it has a time to pass into the parameter. The end date is used to check that the time received from the video is still within the current transcript line.

PR #156, opened 2020-08-10
Removes videoId from URL

jaysaleh:
This pr removes the videoId from the URL


src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -94,7 +94,7 @@ public void doPost_urlAlreadyInDatabase_shouldReturnLecture() throws IOException
     servlet.doPost(request, response);
 
     assertThat(datastoreService.prepare(new Query(LectureUtil.KIND)).countEntities()).isEqualTo(1);
-    verify(response).sendRedirect("/view/?id=1&video-id=wXhTHyIgQ_U");
+    verify(response).sendRedirect("/view/?id=1");
```
amuamushu, line 97:
```suggestion
    verify(response).sendRedirect("/view?id=1");
```
I don't think there should be an extra slash after `view`. I think the issue may be the `REDIRECT_URL` in `LectureServlet`.

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -105,7 +105,7 @@ public void doPost_urlNotInDatabase_shouldAddToDatabaseAndReturnRedirect() throw
...
```
amuamushu, line 108: here too, extra slash
```suggestion
    verify(response).sendRedirect("/view?id=1");
```

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
jaysaleh: fixing in a separate pr

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -152,17 +152,14 @@ private Entity getLectureEntityFromRequest(HttpServletRequest request) {
   }
 
   /**
-   * Returns URL redirecting to lecture view page with parameters {@code lectureId}
-   * and {@code videoId} found in {@code lectureEntity}.
+   * Returns URL redirecting to lecture view page with parameter {@code lectureId}
```
n-wach, line 155: Please explain `lectureId` in the comment, or remove.

I think: `Returns URL for the lecture view page for {@code lectureEntity}` might be good enough

jaysaleh: Fixed

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
n-wach: Technically I think it should work with or without the slash, but yeah we should be consistent and not use one.

jaysaleh:
> Does the transcript stuff work correctly without a video id in the URL? Or is it used somewhere. I notice the window.location.search is still passed as an arg in loadTranscript

Right now the transcripts don't work, was thinking we could do it in a separate PR

src/main/webapp/view/transcript/transcript.js
```diff
@@ -21,6 +21,7 @@ const ENDPOINT_TRANSCRIPT = '/transcript';
  */
 // TODO: Delete this method once TranscriptServlet's doPost()
 // is called in LectureServlet.
+// TODO: Add videoId as a param.
```
billie-cat, line 24: Why add this if the other TODO says to delete the method?

jaysaleh: For transcripts to work while this function exists the param needs to be added

Issue #165, opened 2020-08-11
Fix issue with the Nav bar hiding upon scrolling the transcript lines.


amuamushu:
Currently, when a transcript line is scrolled to, it also scrolls the nav bar out of view.

Potential Solutions:
- Adjust position of nav bar --> Spacing between the nav bar and the rest of the page differs based on whether the transcript is scrolling or not.
- Update the container's `.scrollTop` property based on `element.offsetHeight`

PR #167, opened 2020-08-11
Fix height of transcript container.

amuamushu:
This pull request fixes the issue in which the transcript does not occupy the entire transcript container space.

Screenshot of the full page: https://screenshot.googleplex.com/RftFwDugiMt.png

### DIFF
--- a/src/main/webapp/view/lecture-view.css
+++ b/src/main/webapp/view/lecture-view.css
@@ -66,7 +66,7 @@
 }
 
 #transcript-lines-container {
-  height: 100%;
+  height: 90%;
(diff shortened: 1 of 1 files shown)


amuamushu:
> Looks better, but when I run this locally and the transcript auto scrolls, the top of the page gets cut off. This includes the navbar and most of the video now. Can we get that issue fixed in this PR too?

Thanks for noticing that! I'm working on fixing that issue right now (#165) in a different pull request because it might take a while to look into.

n-wach:
Sounds good, approved!

PR #170, opened 2020-08-11
Automatically scroll discussion to relevant comments

n-wach:
This adds scrolling to seekDiscussion.  Every time seekDiscussion is called, we instantly scroll to the first comment within 10 seconds of the reported current time.


src/main/webapp/view/discussion/discussion.js
```diff
@@ -147,6 +154,30 @@ function updateNewCommentTimestamp(timeMilliseconds) {
...
+/**
+ * Returns an array of the {@code DiscussionComment}s within
+ * {@code TIME_TOLERANCE_MILLISECONDS} milliseconds to the {@code
+ * timestampMilliseconds}. This returns an empty array if no elements are
+ * nearby.
+ */
+function getNearbyDiscussionComments(timestampMilliseconds) {
+  const nearby = [];
+  // currentRootDiscussionComments is already sorted by timestamp.
+  for (const element of currentRootDiscussionComments) {
```
amuamushu, line 166: When we have a lot of comments, I feel it is inefficient to loop through all of the current root comments like so. 

Do you think you can look into using binary search to find the closest element to the lower time bound and start iterating from there instead?

src/main/webapp/view/discussion/discussion.js
```diff
@@ -147,6 +154,30 @@ function updateNewCommentTimestamp(timeMilliseconds) {
   newCommentTimestampMilliseconds = timeMilliseconds;
 }
 
+/**
+ * Returns an array of the {@code DiscussionComment}s within
+ * {@code TIME_TOLERANCE_MILLISECONDS} milliseconds to the {@code
```
amuamushu, line 159: I think we should be using backticks instead of @code (#171)

src/main/webapp/view/discussion/discussion.js
```diff
@@ -147,6 +154,30 @@ function updateNewCommentTimestamp(timeMilliseconds) {
...
+ * timestampMilliseconds}. This returns an empty array if no elements are
```
amuamushu, line 160: nit: Can `This returns an empty ...` be moved to its own paragraph

src/main/webapp/view/discussion/discussion.js
```diff
@@ -36,7 +36,11 @@ const SELECTOR_CANCEL_REPLY = '#cancel-reply';
 const SELECTOR_POST_REPLY = '#post-reply';
 const SELECTOR_REPLY_TEXTAREA = '#reply-textarea';
 
+// 10 seconds.
+const TIME_TOLERANCE_MILLISECONDS = 10 * 1000;
+
 let newCommentTimestampMilliseconds = 0;
+let currentRootDiscussionComments = [];
```
amuamushu, line 43: Can you add an inline comment for the array type? 
https://google.github.io/styleguide/jsguide.html#features-declare-types-as-needed

src/main/webapp/view/discussion/discussion.js
n-wach: I just tested with 1000 comments (our goal in design doc) at timestamp 00:00.  

It takes 1.5 milliseconds to run `getNearbyDiscussionComments(0)` and return an array with them all.  So this should be fine, especially as usually the comments wont all be grouped at 00:00 and only a few would be added to the array.

src/main/webapp/view/discussion/discussion.js
```diff
@@ -36,7 +36,11 @@ const SELECTOR_CANCEL_REPLY = '#cancel-reply';
 const SELECTOR_POST_REPLY = '#post-reply';
 const SELECTOR_REPLY_TEXTAREA = '#reply-textarea';
 
+// 10 seconds.
+const TIME_TOLERANCE_MILLISECONDS = 10 * 1000;
```
jaysaleh, line 40: Can you change this to 10,000 and leave the comment

src/main/webapp/view/discussion/discussion.js
```diff
@@ -91,11 +95,14 @@ async function postAndReload(
 async function loadDiscussion() {
   // Clear any existing comments before loading.
   ELEMENT_DISCUSSION.textContent = '';
+  currentRootDiscussionComments = [];
 
   const comments = await fetchDiscussion();
   const preparedComments = prepareComments(comments);
   for (const comment of preparedComments) {
-    ELEMENT_DISCUSSION.appendChild(new DiscussionComment(comment));
+    const element = new DiscussionComment(comment);
```
jaysaleh, line 103: This name could be more descriptive element is general

src/main/webapp/view/discussion/discussion.js
n-wach: Done

src/main/webapp/view/discussion/discussion.js
n-wach: Done

src/main/webapp/view/discussion/discussion.js
n-wach: Added!

src/main/webapp/view/discussion/discussion.js
n-wach: changed to `commentElement`

src/main/webapp/view/discussion/discussion.js
n-wach: Done!

src/main/webapp/view/discussion/discussion.js
```diff
@@ -276,7 +276,7 @@ class DiscussionComment extends HTMLElement {
 // Custom element names must contain a hyphen.
 customElements.define('discussion-comment', DiscussionComment);
 
-/** Seeks discussion to {@code currentTime}. */
+/** Seeks discussion to `@code currentTime`. */
```
amuamushu, line 279:
```suggestion
/** Seeks discussion to `currentTimeSeconds`. */
```

n-wach: ah yes good catch

src/main/webapp/view/discussion/discussion.js
amuamushu: Sounds good. Thanks for looking into that!

src/main/webapp/view/discussion/discussion.js
n-wach: Fixed!

src/main/webapp/view/discussion/discussion.js
```diff
@@ -147,6 +154,30 @@ function updateNewCommentTimestamp(timeMs) {
   newCommentTimestampMs = timeMs;
 }
 
+/**
+ * Returns an array of the `DiscussionComment`s within
+ * `TIME_TOLERANCE_MILLISECONDS` milliseconds to the `timestampMilliseconds`.
```
billie-cat, line 159: why are these all quoted? Shouldn't you use @code instead?

src/main/webapp/view/discussion/discussion.js
```diff
@@ -147,6 +154,30 @@ function updateNewCommentTimestamp(timeMs) {
...
+ *
+ * <p>This returns an empty array if no elements are nearby.
```
billie-cat, line 161: This is important information and should be part of the main JSDoc, not in a new paragraph.

src/main/webapp/view/discussion/discussion.js
```diff
@@ -147,6 +154,30 @@ function updateNewCommentTimestamp(timeMs) {
...
+  // currentRootDiscussionComments is already sorted by timestamp.
+  for (const element of currentRootDiscussionComments) {
+    const commentTime = element.comment.timestampMilliseconds;
+    if (commentTime < timestampMilliseconds - TIME_TOLERANCE_MILLISECONDS) {
+      // Before the start of the range, continue to next.
+      continue;
+    }
+    if (commentTime > timestampMilliseconds + TIME_TOLERANCE_MILLISECONDS) {
+      // Outside of range, there will be no more.
+      break;
```
billie-cat, line 174: Instead of break, you can return nearby.

src/main/webapp/view/discussion/discussion.js
```diff
@@ -230,14 +261,29 @@ class DiscussionComment extends HTMLElement {
     span.slot = name;
     this.appendChild(span);
   }
+
+  /**
+   * Scroll `ELEMENT_DISCUSSION` such that this element is at the top.
```
billie-cat, line 266: why quote it instead of using @code?

billie-cat: Also you are talking about the code and someone needs to look up what ELEMENT_DISCUSSION is. Why not just say: Scroll the discussion element to the first discussion created. Or something like that.

src/main/webapp/view/discussion/discussion.js
```diff
@@ -36,7 +36,11 @@ const SELECTOR_CANCEL_REPLY = '#cancel-reply';
 const SELECTOR_POST_REPLY = '#post-reply';
 const SELECTOR_REPLY_TEXTAREA = '#reply-textarea';
 
+// 10 seconds.
+const TIME_TOLERANCE_MILLISECONDS = 10000;
+
 let newCommentTimestampMs = 0;
+let /** !Array<DiscussionComment> */ currentRootDiscussionComments = [];
```
f-camp, line 43: Why do we have an inline comment here please?

src/main/webapp/view/discussion/discussion.js
f-camp: @billie-cat I believe the javascript style prefers ticks to @code, this came up in yesterday sync.

f-camp, line 159: Perhaps this can be rephrased to

Return an array of `DiscussionComment`s in the range of  `TIME_TOLERANCE_MILLISECONDS` milliseconds and `timestampMilliseconds`?

It's a bit unclear the way it currently is.

src/main/webapp/view/discussion/discussion.js
f-camp: ditto re: @code vs ticks

f-camp: What's ```ELEMENT_DISCUSSION``` sorry? is this the current discussion element in the ui? If so perhaps what we're looking for is 
"Scroll such that this element is at the top"?

src/main/webapp/view/discussion/discussion.js
n-wach: See: #171

src/main/webapp/view/discussion/discussion.js
n-wach: Type annotation for unclear variable type.  See: https://google.github.io/styleguide/jsguide.html#features-declare-types-as-needed

src/main/webapp/view/discussion/discussion.js
n-wach: Combined

src/main/webapp/view/discussion/discussion.js
n-wach: Done

src/main/webapp/view/discussion/discussion.js
n-wach: "Scroll such that this element is at the top of the discussion area."

src/main/webapp/view/discussion/discussion.js
n-wach: I moved things around, let me know how it looks

PR #174, opened 2020-08-11
Update seekTranscript to use long instead of Date for time.

amuamushu:
Currently, seekTranscript uses Date. However, a recent pull request updated the rest of the code to use long.


src/main/webapp/timestamps.js
```diff
@@ -42,11 +41,8 @@ function timestampToSeconds(timestampMs) {
...
 /**
- * @return {number} The {@code date} in seconds.
+ * Converts `seconds` into milliseconds.
  */
-// TODO: Remove this method once the pull request updating
-// Date to long is approved.
-function getDateInSeconds(date) {
-  return date.getSeconds() + date.getMinutes() * MINUTES_TO_SECONDS +
-      date.getHours() * HOURS_TO_SECONDS;
+function secondsToMilliseconds(seconds) {
```
jaysaleh, line 46: nit: maybe rename `seconds` we already know seconds is being passes in by the function name

src/main/webapp/view/transcript/transcript.js
```diff
@@ -88,8 +88,8 @@ function appendTextToList(transcriptLine, ulElement) {
   liElement.appendChild(hrElement);
   ulElement.appendChild(liElement);
   // Save the dates for seeking later.
```
n-wach, line 90: update/remove this comment

amuamushu: Thanks for reminding me! Removed.

src/main/webapp/timestamps.js
amuamushu: I think I'm going to keep it as `seconds` if that's okay because I want the function to be general.

billie-cat, line 46: wasn't this code added by Nathan? Should this PR be synced?

amuamushu: No, this was a function I added. His function converts milliseconds to seconds (the other way around)

PR #176, opened 2020-08-12
Update Scroll method for the Transcript.

amuamushu:
Currently, when a transcript line is scrolled to, it also scrolls the nav bar out of view. This pull request updates the scrolling function to prevent that from happening.

transcript scrolling with nav bar still there: https://drive.google.com/file/d/1aqYlpVaF4iGHttQf-nrdVyZBBBNbMMgp/view?usp=sharing


src/main/webapp/view/transcript/transcript.js
```diff
@@ -142,3 +142,13 @@ function removeBold(transcriptLineLiElement) {
   transcriptLineLiElement.classList.add(DEFAULT_FONT_WEIGHT);
   transcriptLineLiElement.classList.remove(BOLD_FONT_WEIGHT);
 }
+
+/**
+ * Scrolls `transcriptLine` to the top of
+ * `TRANSCRIPT_CONTAINER`.
```
n-wach, line 148: It's unclear what `TRANSCRIPT_CONTAINER` is without reading other code.  Change to "Scrolls `transcriptLine` to the top of the transcript area."?

amuamushu: Gotcha, updated!

Issue #178, opened 2020-08-12
Verify presence of required servlet parameters


n-wach:
Many servlets will throw errors or use default values when certain parameters are not supplied.  We want to check the params are supplied, and use `.sendError` if they are not.

This affects:
- [x] Discussion Servlet
- [x] Transcript Servlet
- [x] Lecture Servlet

These should be done in separate PRs

PR #184, opened 2020-08-12
Require parameters in DiscussionServlet

n-wach:
This addresses requiring params in the discussion servlet for #178 

Note: this doesn't check the validity of parameters (like numbers are numbers), only the presence.  IMHO we can just let an internal server error leak out for now if that's the case.

Next will be lecture servlet.


src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -63,24 +63,45 @@ public void init() throws ServletException {
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    if (request.getParameter(PARAM_LECTURE) == null) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Missing lecture parameter.");
```
amuamushu, line 67: nit: can you create constants for the error messages?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -88,6 +88,55 @@ public void doPost_loggedOut_postForbidden() throws ServletException, IOExceptio
         .isEqualTo(0);
   }
 
+  @Test
+  public void doPost_missingLecture_badRequest() throws ServletException, IOException {
+    testServices.setEnvIsLoggedIn(true);
+
+    servlet.doPost(request, response);
+
+    verify(response).sendError(HttpServletResponse.SC_BAD_REQUEST, "Missing lecture parameter.");
```
amuamushu, line 97: please create constants for the error messages

n-wach: Done!

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: Done!

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -52,6 +52,15 @@
   @VisibleForTesting static final String PARAM_TIMESTAMP = "timestamp";
   @VisibleForTesting static final String PARAM_TYPE = "type";
 
+  @VisibleForTesting static final String ERROR_MISSING_LECTURE = "Missing lecture parameter.";
+  @VisibleForTesting
+  static final String ERROR_MISSING_COMMENT_TYPE = "Missing comment type parameter.";
```
jaysaleh, line 57: Did the linter cause this?

n-wach: Yeah must have... I'll change to `ERROR_MISSING_TYPE` so it fits on a line

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -83,11 +83,62 @@ public void doPost_loggedOut_postForbidden() throws ServletException, IOExceptio
...
+
+    verify(response).sendError(
+        HttpServletResponse.SC_BAD_REQUEST, DiscussionServlet.ERROR_MISSING_LECTURE);
+  }
+
+  @Test
+  public void doPost_missingType_badRequest() throws ServletException, IOException {
+    testServices.setEnvIsLoggedIn(true);
+    when(request.getParameter(DiscussionServlet.PARAM_LECTURE)).thenReturn(LECTURE_ID_STR);
+    when(request.getReader()).thenReturn(new BufferedReader(new StringReader("Untested Content")));
```
jaysaleh, line 106: "Untested Content" is used pretty frequently, consider declaring it global

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: Actually nvm, that wouldn't cause it to fit

billie-cat:
Concerning: Note: this doesn't check the validity of parameters (like numbers are numbers), only the presence. IMHO we can just let an internal server error leak out for now if that's the case.

Do you have a bug to track this?

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
billie-cat: I wouldn't do that as mentioned in the "Local Variables and Class-Wide Constants" section of the unit test tech resources (https://docs.google.com/document/d/18T9ef-DRqTncVD377GoIQmOGI84cNKkZ7Nn07YURFTw/edit#heading=h.x20ltbjso2oq)

n-wach:
> Concerning: Note: this doesn't check the validity of parameters (like numbers are numbers), only the presence. IMHO we can just let an internal server error leak out for now if that's the case.
> 
> Do you have a bug to track this?

Created #199

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -52,6 +52,15 @@
...
+  @VisibleForTesting
+  static final String ERROR_MISSING_PARENT = "Missing parent parameter for reply comment.";
+  @VisibleForTesting
+  static final String ERROR_MISSING_TIMESTAMP = "Missing timestamp parameter for root comment.";
+  @VisibleForTesting static final String ERROR_NOT_LOGGED_IN = "You are not logged in.";
```
f-camp, line 62: Concerning @VisibleForTesting, for static fields it makes more sense to have these as public and call them directly in tests. In general @VisibleForTesting should be last resort after considering other options.

n-wach: Should PARAMs be public too?  I thought we wanted to make them the most protected as possible, so used @VisibleForTesting because they can't be private

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -63,24 +72,43 @@ public void init() throws ServletException {
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    if (request.getParameter(PARAM_LECTURE) == null) {
```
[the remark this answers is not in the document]
n-wach: Had chat and created helper

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -94,6 +122,10 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
+    if (request.getParameter(PARAM_LECTURE) == null) {
```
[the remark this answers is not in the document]
n-wach: datastore was fetched in `init`, removed the get here--should have been removed long ago

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: "If you expose constants in Java, please mark the modifications with @VisibleForTesting"
This is from the doc: https://docs.google.com/document/d/18T9ef-DRqTncVD377GoIQmOGI84cNKkZ7Nn07YURFTw/edit

f-camp: https://docs.google.com/document/d/1ogapKfwinMeWHbJxMHzjrGxpoAiz-r736CXQ8y5CPxY/edit#heading=h.fmos5zqp4q3d

I think the use of this is really a balance. A lot of the time it signifies a code smell if one needs to do this.
Maybe the right thing here is to move this strings to a Util  file like we had for the other Servlets. If many strings now need to be shared outside of this class they should be in a DiscussionUtil class under /utils like the other Servlets.

billie-cat: I don't think making visible for testing is so bad. This is only exposing a variable, not a method. I don't think it makes sense to move it to a Util because they are only used in this class. If you do not want to make @VisibleForTesting, then when testing the error message just pass the string directly in. It is not common practice to create const variables for error messages. Just do:

assertThat(response.errorMessage).isEqualTo("error message").

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -63,19 +73,26 @@ public void init() throws ServletException {
...
+    }
+
     User author = userService.getCurrentUser();
     if (author == null) {
-      response.sendError(HttpServletResponse.SC_FORBIDDEN, "You are not logged in.");
+      response.sendError(HttpServletResponse.SC_FORBIDDEN, ERROR_NOT_LOGGED_IN);
       return;
     }
+
+    long lectureId = Long.parseLong(request.getParameter(PARAM_LECTURE));
```
billie-cat, line 88: Order of declaration is very important for readability of a method. I think this can be improved. I would declare in this order (obviously this is not a hard rule).

final Entity commentEntity; // This is the first used variable at line 99
long lectureId ... // used to create lecture which is second var.
Key lecture ... // Second used var
String content ... // Third if you ignore parent and author.
Date dateNow ... // Last
< new line>
Comment.Type type = ... // I declare this right before the if statement so it is obvious to the user where it is used (slightly more important than your other declarations and therefore split from the rest)
if (type == ...)


You can read more about it here: 4.8.2.2 Declared when needed
Local variables are not habitually declared at the start of their containing block or block-like construct. Instead, local variables are declared close to the point they are first used (within reason), to minimize their scope. Local variable declarations typically have initializers, or are initialized immediately after declaration. 

https://google.github.io/styleguide/javaguide.html#s4.8.2-variable-declarations

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -85,14 +102,41 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  private Optional<String> validatePostRequest(HttpServletRequest request) {
+    if (request.getParameter(PARAM_LECTURE) == null) {
+      return Optional.of(ERROR_MISSING_LECTURE);
+    }
+    if (request.getParameter(PARAM_TYPE) == null) {
+      return Optional.of(ERROR_MISSING_COMMENT_TYPE);
+    }
+
+    Comment.Type type = Comment.Type.valueOf(request.getParameter(PARAM_TYPE));
+    if (type == Comment.Type.REPLY) {
```
billie-cat, line 119: These nested if statements are not necessary. When you see nested if-statements you should try to re-write it to something simpler. 

if (type == Comment.Type.REPLY && request.getParameter(PARAM_PARENT) == null) {
  return error;
 }

if (request.getParameter(PARAM_TIMESTAMP) == null) {
  return error;
}

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -79,15 +79,72 @@ public void cleanUp() {
   @Test
   public void doPost_loggedOut_postForbidden() throws ServletException, IOException {
     testServices.setEnvIsLoggedIn(false);
+    final int parentId = 32;
```
billie-cat, line 82: remove this as it adds complexity for no reason. This variable is used one and converted to a string. Just pass the string directly.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
billie-cat: So instead of making "Untested Content" a class-wide const I would make the BufferReader a class-wide variable. 

For simple types prefer raw values (ie, 1, null, Optional.empty(), "string"), but this BufferReader is not that simple so it can become a class-wide variable.

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
```diff
@@ -118,8 +175,7 @@ public void doPost_reply_storesCommentWithAllProperties() throws ServletExceptio
   public void doPost_rootComment_storesTypeAndTimestamp() throws ServletException, IOException {
     testServices.setEnvIsLoggedIn(true);
     when(request.getParameter(DiscussionServlet.PARAM_LECTURE)).thenReturn(LECTURE_ID_STR);
-    when(request.getParameter(DiscussionServlet.PARAM_TIMESTAMP))
-        .thenReturn(TIMESTAMP_MS_STRING);
+    when(request.getParameter(DiscussionServlet.PARAM_TIMESTAMP)).thenReturn(TIMESTAMP_MS_STRING);
```
billie-cat, line 178: TIMESTAMP_MS_STRING and LECTURE_ID_STR add complexity. It is much easier to read the following:

when(request.getParameter(DiscussionServlet.PARAM_LECTURE)).thenReturn("1");
    when(request.getParameter(DiscussionServlet.PARAM_TIMESTAMP))	    when(request.getParameter(DiscussionServlet.PARAM_TIMESTAMP)).thenReturn("2000");

assertThat(comment.timestampMs().get()).isEqualTo(2000);

It is easy to see that "2000" and 2000 are the same. TIMESTAMP_MS_STRING and TIMESTAMP_MS are 2 different variables and therefore I need to look at their declaration to understand how they are related.

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
f-camp: Spoke to @billie-cat  offline. The  @VisibleForTesting usage is truly a balance and if there are too many in a file its definitely a code smell. I'm ok with them being  inline in the assertions.

Please remove @VisbleForTesting and check inline like @billie-cat has mentioned.
```assertThat(response.errorMessage).isEqualTo("error message")```.

In general if a string is being used in one place in a test lets start making those inline to prevent this confusion.

File: src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java (hunk 8 above)
Comment (n-wach) [reply src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java:89]: Reordered as you suggested

File: src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java (hunk 9 above)
Comment (n-wach) [reply src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java]: Refactored

File: src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java (hunk 10 above)
Comment (n-wach) [reply src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java:82]: It was also used in the assertion to check the value was set.  Removed in favor of 2 literals

File: src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java (hunk 11 above)
Comment (n-wach) [reply src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java:178]: Right.  Removed!

File: src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java [hunk 12]
```diff
@@ -131,11 +178,19 @@ public void doPost_rootComment_storesTypeAndTimestamp() throws ServletException,
...
     assertThat(comment.type()).isEqualTo(Comment.Type.QUESTION);
   }
 
+  @Test
+  public void doGet_missingLecture_badRequest() throws ServletException, IOException {
+    testServices.setEnvIsLoggedIn(true);
+
+    servlet.doGet(request, response);
+
+    verify(response).sendError(HttpServletResponse.SC_BAD_REQUEST, "Missing lecture parameter.");
```
billie-cat, line 191: For these raw errors you should add the argument comment /* =*/

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: Done in all 3 PRs!  Thanks for figuring out the best way to do this :)

src/test/java/com/googleinterns/zoomtube/servlets/DiscussionServletTest.java
n-wach: Added for all 3 PRs

PR #187, opened 2020-08-12
Require parameters in TranscriptServlet

n-wach:
Last PR for #178 

This one is short--only one parameter is used.

### DIFF
--- a/src/main/java/com/googleinterns/zoomtube/servlets/TranscriptServlet.java
+++ b/src/main/java/com/googleinterns/zoomtube/servlets/TranscriptServlet.java
@@ -63,14 +63,23 @@ public class TranscriptServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    if (request.getParameter(PARAM_ID) == null) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/googleinterns/zoomtube/servlets/TranscriptServlet.java
```diff
@@ -53,6 +53,7 @@
  */
 public class TranscriptServlet extends HttpServlet {
   private DatastoreService datastore;
+  @VisibleForTesting static final String PARAM_ID = "id";
```
amuamushu, line 56: I think it's better to use `LectureUtil.ID` because the id is based off the redirect Url in `LectureServlet`. If `LectureUtil.ID` changes, `PARAM_ID` would also have to change.

n-wach: Thanks for pointing that out!

This value is now used to generate the queries in javascript, so I think it makes sense to respecify it in each servlet that'll be handling a request with it.

n-wach: Refactored

n-wach:
Addressed, please see other PR for a link to best-practices explaining why VisibleForTesting is justified here

PR #192, opened 2020-08-12
Convert the Transcript Line on the DOM to be a Custom Element.

amuamushu:
This change refactors the methods that involving a transcriptLine ul element by moving them inside the `TranscriptLine` Javascript class.


src/main/webapp/timestamps.js
```diff
@@ -39,3 +39,13 @@ function timestampToString(timestampMs) {
 function secondsToMilliseconds(seconds) {
   return Math.round(seconds * MILLISECONDS_PER_SECOND);
 }
+
+/**
+ * Returns {String} a time range from
+ * `startTimestampMs` to `endTimestampMs`.
+ */
```
amuamushu, line 46: This function is patched from pull request #228 . Please ignore this function for this pull request. Thank you!

src/main/webapp/view/transcript/transcript.js
```diff
@@ -115,60 +86,108 @@ function seekTranscript(currentTime) {
...
- * is currently bolded.
- */
-function removeBold(transcriptLineLiElement) {
-  if (!isBolded(transcriptLineLiElement)) {
-    return;
+  /**
+   * Updates the template slot `slotName` with `slotValue` and styling
+   * from `slotStyle`.
+   */
+  updateSlot(slotName, slotValue, slotStyle) {
```
jaysaleh, line 137: nit: maybe rename this to `updateTemplateSlot`

amuamushu: done, thanks!

src/main/webapp/view/transcript/transcript.js
```diff
@@ -115,60 +86,109 @@ function seekTranscript(currentTime) {
...
   }
-  transcriptLineLiElement.classList.add(DEFAULT_FONT_WEIGHT);
-  transcriptLineLiElement.classList.remove(BOLD_FONT_WEIGHT);
-}
 
-/** Returns true if`transcriptLineLiElement` is bolded. */
-function isBolded(transcriptLineLiElement) {
-  return transcriptLineLiElement.classList.contains(BOLD_FONT_WEIGHT);
-}
+  /** Returns true if the text is already bolded. */
```
n-wach, line 146: Change to `if this element is already bolded`?

src/main/webapp/view/transcript/transcript.js
```diff
@@ -115,60 +86,109 @@ function seekTranscript(currentTime) {
...
- * is currently bolded.
- */
-function removeBold(transcriptLineLiElement) {
-  if (!isBolded(transcriptLineLiElement)) {
-    return;
+  /**
+   * Updates the template slot `slotName` with `slotValue` and styling
+   * from `slotStyle`.
+   */
+  updateTemplateSlot(slotName, slotValue, slotStyle) {
```
n-wach, line 138: Can the `slotStyle` be done in the template HTML code, and removed from this function?  Since it looks like it's always the same class per slot

src/main/webapp/view/transcript/transcript.js
```diff
@@ -115,60 +86,109 @@ function seekTranscript(currentTime) {
...
+   * Removes bold from the text if it
+   * is currently bolded.
+   */
+  removeBold() {
+    if (!this.isBolded()) {
+      return;
+    }
+    this.classList.add(DEFAULT_FONT_WEIGHT);
+    this.classList.remove(BOLD_FONT_WEIGHT);
+  }
```
n-wach, line 173: Now that you're using classes, I feel like it would simplify things if `isBolded`, `addBold`, `removeBold`, were combined into a single `setBold(boolean)`.

Or even a `bold` setter and getter that adds/removes the styling and updates an internal state.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -115,60 +86,109 @@ function seekTranscript(currentTime) {
...
+    return this.startTimestampMs <= currentTimeMs &&
+        currentTimeMs <= this.endTimestampMs;
+  }
+
+  /**
+   * Scrolls this transcript line to the top of the transcript area.
+   */
+  scrollToTopOfTranscript() {
+    const transcriptContainer = document.getElementById(TRANSCRIPT_CONTAINER);
+    const ulElementOffset = this.parentElement.offsetTop;
```
n-wach, line 189: Is `this.parentElement` the same as `transcriptContainer`?

amuamushu: No, it's the ul tag that the transcript-line tags are nested under. `TranscriptContainer` contains the ul tag. The nesting of the tags is a bit weird though. I created an issue to fix this later (#236).

src/main/webapp/view/transcript/transcript.js
n-wach: Ah, sounds good!

src/main/webapp/view/transcript/transcript.js
n-wach: I was just wondering, nbd

src/main/webapp/view/transcript/transcript.js
n-wach: Maybe you can wrap the `<slot>` in a `<span>` with the correct styling?  I know it's sometimes unavoidable but I think it's best practice to try to separate styling (especially static stuff like this) from code as much as possible

amuamushu: Gotcha, wrapping it works! Thanks!

src/main/webapp/view/transcript/transcript.js
```diff
@@ -47,45 +51,13 @@ function addMultipleTranscriptLinesToDom(transcriptLines) {
...
-  // Sets the current transcript line to be the first line.
-  if (currentTranscriptLine == null) {
-    currentTranscriptLine = liElement;
-  }
-}
-
-/**
- * Creates a p tag to store the given `text` inside the
- * `container`.
+ * Creates a p tag to store the given {@code text} inside the
```
billie-cat, line 59: Should this @code be removed?

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +85,106 @@ function seekTranscript(currentTime) {
...
+  currentTranscriptLine.scrollToTopOfTranscript();
+  currentTranscriptLine.addBold();
   // TODO: Handle the case where the video isn't only playing.
 }
 
 /**
- * Bolds the text in `transcriptLineLiElement` if it is not already
- * bolded.
+ * Creates a transcript line element containing the text,
+ * start time, and end time from `transcriptLine`.
```
billie-cat, line 101: you are referencing transcriptLine but this is not related to the class. Please keep this generic for the class.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +85,106 @@ function seekTranscript(currentTime) {
...
+   * Creates a custom HTML element representing a transcript line.
+   *
+   * <p>Uses the template and slots defined in `TRANSCRIPT_TEMPLATE` to
+   * help create the transcript line.
+   *
+   * @param transcriptLine The transcriptLine from `ENDPOINT_TRANSCRIPT`
+   *     whose `attributes` should be used.
+   */
+  constructor(transcriptLine) {
+    super();
```
billie-cat, line 114: It is best practise to not do work in the constructor. Please see: https://docs.google.com/document/d/14CEPF-d3eF6HzG4AhviQiRbdEaVF8_BDOwKHMUmqFEQ/edit

I would use a factory method to create the object.

n-wach: This is the same situation as in the DiscussionComment class, which we talked about and okay-ed.  Both use templates and need to initialize various things before they are rendered.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +85,106 @@ function seekTranscript(currentTime) {
...
+    this.attachShadow({mode: 'open'});
+    this.shadowRoot.appendChild(template.content.cloneNode(true));
+    this.updateTemplateSlot(TRANSCRIPT_SLOT_TIME_RANGE, timestampRange);
+    this.updateTemplateSlot(TRANSCRIPT_SLOT_CONTENT, transcriptLine.content);
+
+    this.startTimestampMs = transcriptLine.startTimestampMs;
+    this.endTimestampMs = transcriptLine.endTimestampMs;
+    // Sets the current transcript line to be the first line.
+    if (currentTranscriptLine == null) {
+      currentTranscriptLine = this;
```
n-wach, line 111: This is an unclear side effect that should probably be moved out of the constructor

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +85,106 @@ function seekTranscript(currentTime) {
...
+        transcriptLine.startTimestampMs, transcriptLine.endTimestampMs);
+
+    const template = document.getElementById(TRANSCRIPT_TEMPLATE);
+    this.attachShadow({mode: 'open'});
+    this.shadowRoot.appendChild(template.content.cloneNode(true));
+    this.updateTemplateSlot(TRANSCRIPT_SLOT_TIME_RANGE, timestampRange);
+    this.updateTemplateSlot(TRANSCRIPT_SLOT_CONTENT, transcriptLine.content);
+
+    this.startTimestampMs = transcriptLine.startTimestampMs;
+    this.endTimestampMs = transcriptLine.endTimestampMs;
```
n-wach, line 108: It might be better to just save everything with `this.transcriptLine = transcriptLine` instead of copying over the various values you'll need later.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +85,106 @@ function seekTranscript(currentTime) {
...
+    }
+    this.classList.add(DEFAULT_FONT_WEIGHT);
+    this.classList.remove(BOLD_FONT_WEIGHT);
+  }
+
+  /**
+   * Returns true if `currentTimeMs` is within the time range for
+   * this transcript line.
+   */
+  isWithinCurrentTimeRange(currentTimeMs) {
```
n-wach, line 158: Can you change method and argument name to be more generic?

src/main/webapp/view/transcript/transcript.js
amuamushu: True, I deleted this if statement for now and added a TODO for making this an instance attribute when `transcript.js` is refactored into a class. This should move the responsibility of keeping track of the current transcript line outside of the `transcriptLine` class and to where `seekTranscript()` is. Thanks!

src/main/webapp/view/transcript/transcript.js
amuamushu: Thanks for catching that! I must have missed it in this branch.

src/main/webapp/view/transcript/transcript.js
amuamushu: gotcha, I remove the last bit (`from transcriptLine`) from the description.

src/main/webapp/view/transcript/transcript.js
amuamushu: Yes, thank you for catching that!

src/main/webapp/view/transcript/transcript.js
billie-cat: I ended up talking to my co-worker about this and independent of the language, doing work in the constructor is not a good practise. It is best to create a static method in the class which does all the work and passes the fields to the constructor.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -13,12 +13,18 @@
...
 const DEFAULT_FONT_WEIGHT = 'text-muted';
 const BOLD_FONT_WEIGHT = 'font-weight-bold';
-const PARAM_ID = 'id';
+const URL_PARAM_ID = 'id';
+const TRANSCRIPT_SLOT_TIME_RANGE = 'timestamp-range';
+const TRANSCRIPT_SLOT_CONTENT = 'content';
+const CUSTOM_ELEMENT_TRANSCRIPT_LINE = 'transcript-line';
 
 let /** Element */ currentTranscriptLine;
+// TODO: Create an instance reference to the current transcriptLine when
```
billie-cat, line 26: should "current transcriptLine be currentTranscriptLine?

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +70,101 @@ function seekTranscript(currentTime) {
...
 /**
- * Bolds the text in `transcriptLineLiElement` if it is not already
- * bolded.
+ * Creates a transcript line element containing the text,
+ * start time, and end time.
  */
-function addBold(transcriptLineLiElement) {
-  if (isBolded(transcriptLineLiElement)) {
-    return;
+class TranscriptLine extends HTMLElement {
```
billie-cat, line 88: TranscriptLine is slightly generic class name? What will your other class names be for TranscriptLine related code? I thought you were going to name this TranscriptLineElement and then have another class be TranscriptLineSeeker (or something like that).

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +70,101 @@ function seekTranscript(currentTime) {
...
+  /**
+   * Creates a custom HTML element representing a transcript line.
```
billie-cat, line 90:
```suggestion
   * Creates a custom HTML element representing `transcriptLine`.
```

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +70,101 @@ function seekTranscript(currentTime) {
...
- * Returns true if `currentTimeMs` is within the time range for
- * the current transcript line.
- */
-function isWithinCurrentTimeRange(currentTimeMs) {
-  return currentTranscriptLine.startTimestampMs <= currentTimeMs &&
-      currentTimeMs <= currentTranscriptLine.endTimestampMs;
-}
+  /**
+   * Bolds the element if it is not already
+   * bolded.
```
billie-cat, line 129: nit: the new line for the work bolded seems a bit weird. You should only add new lines if it goes above a certain character length.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +70,101 @@ function seekTranscript(currentTime) {
...
+    }
+    this.classList.add(DEFAULT_FONT_WEIGHT);
+    this.classList.remove(BOLD_FONT_WEIGHT);
+  }
+
+  /**
+   * Returns true if `timeMs` is within the time range for
+   * this transcript line.
+   */
+  isWithinTimeRange(timeMs) {
```
billie-cat, line 155: nit: is timeMs a timestamp? should it be timestampMs?

src/main/webapp/view/transcript/transcript.js
```diff
@@ -113,60 +70,101 @@ function seekTranscript(currentTime) {
...
+    return this.transcriptLine.startTimestampMs <= timeMs &&
+    timeMs <= this.transcriptLine.endTimestampMs;
+  }
+
+  /**
+   * Scrolls this transcript line to the top of the transcript area.
+   */
+  scrollToTopOfTranscript() {
```
billie-cat, line 163: if this class is about a TranscriptLineElement, then how can it scroll to the top of the transcript? That would implement multiple transcript lines.

src/main/webapp/view/transcript/transcript.js
f-camp: The part I have a problem with in this ```createTimestampRange()``` That seems to me the part that is doing some work. If we can remove that and put it in some kind of initialization method that would work for me.

amuamushu: (Resolved in chat) I kept the code for attaching the shadow in the constructor but moved the code for creating the timerange and updating the slots to a static method.

src/main/webapp/timestamps.js
billie-cat: Will you remove this before merging this PR?

amuamushu: Yes :D

PR #198, opened 2020-08-13
Fix bug in which the first transcript line is not highlighted.

amuamushu:
The first transcript line is never bolded because the styling of a transcript line is only updated when the current transcript line is updated to be the next line.
This PR fixes that issue by adding additional checks in `seekTranscript`

### DIFF
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -126,7 +126,7 @@ function seekTranscript(currentTime) {
   // TODO: Handle the case where the video isn't only playing.
 }
 
-/** 
+/**
  * Bolds the text in `transcriptLineLiElement` if it is not already
  * bolded.
  */
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -115,8 +115,7 @@ function seekTranscript(currentTime) {
   if (currentTimeMs < currentTranscriptLine.startTimestampMs) {
     return;
   }
-  if (currentTranscriptLine.startTimestampMs <= currentTimeMs &&
-      currentTimeMs <= currentTranscriptLine.endTimestampMs) {
+  if (isWithinCurrentTimeRange(currentTimeMs)) {
     if (!isBolded(currentTranscriptLine)) {
       addBold(currentTranscriptLine);
     }
@@ -146,9 +145,13 @@ function isBolded(transcriptLineLiElement) {
   return transcriptLineLiElement.classList.contains(BOLD_FONT_WEIGHT);
 }
 
+/**
+ * Checks if `currentTimeMs` is within the time range for
+ * the current transcript line.
+ * */
 function isWithinCurrentTimeRange(currentTimeMs) {
   return currentTranscriptLine.startTimestampMs <= currentTimeMs &&
-  currentTimeMs <= currentTranscriptLine.endTimestampMs;
+      currentTimeMs <= currentTranscriptLine.endTimestampMs;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/webapp/view/transcript/transcript.js
```diff
@@ -112,7 +112,13 @@ function deleteTranscript() {
 /** Seeks transcript to {@code currentTime}, which is given in seconds. */
 function seekTranscript(currentTime) {
   const currentTimeMs = window.secondsToMilliseconds(currentTime);
-  if (currentTimeMs <= currentTranscriptLine.endTimestampMs) {
+  if (currentTimeMs < currentTranscriptLine.startTimestampMs) {
```
n-wach, line 115: Will `currentTranscriptLine` be `undefined` at first?

n-wach: .

amuamushu: I'm guessing it would be undefined if seekTranscript is called before the transcript finishes loading. 
But once the lines are added to the DOM, it shouldn't be undefined because currentTranscript is assigned when the lines are added to the DOM. 

https://github.com/googleinterns/zoomtube/blob/a8387b467ebd12eff34340f1d6349833d9be4053/src/main/webapp/view/transcript/transcript.js#L83

n-wach: Ok, thanks for adding an issue to track this

amuamushu: Of course, thanks for bring this up! For reference, the issue created is : #211

src/main/webapp/view/transcript/transcript.js
```diff
@@ -112,7 +112,13 @@ function deleteTranscript() {
...
+    return;
+  }
+  if (isWithinCurrentTimeRange(currentTimeMs)) {
+    if (!isBolded(currentTranscriptLine)) {
+      addBold(currentTranscriptLine);
```
billie-cat, line 120: This addBold method could check if the line is already bold and return early if it is. The benefit is that you do not need to have nested if-statements.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -134,6 +140,20 @@ function removeBold(transcriptLineLiElement) {
   transcriptLineLiElement.classList.remove(BOLD_FONT_WEIGHT);
 }
 
+/** Checks if `transcriptLineLiElement` is bolded. */
```
billie-cat, line 143: nit: Reword this to: Returns `true` if ...

The statement: Checks does not imply it returns anything.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -134,6 +140,20 @@ function removeBold(transcriptLineLiElement) {
...
+function isBolded(transcriptLineLiElement) {
+  return transcriptLineLiElement.classList.contains(BOLD_FONT_WEIGHT);
+}
+
+/**
+ * Checks if `currentTimeMs` is within the time range for
```
billie-cat, line 149: here too

src/main/webapp/view/transcript/transcript.js
amuamushu: done! I also added a similar check to `removeBold`.

src/main/webapp/view/transcript/transcript.js
amuamushu: gotcha, done!

src/main/webapp/view/transcript/transcript.js
amuamushu: done!

src/main/webapp/view/transcript/transcript.js
```diff
@@ -124,16 +128,36 @@ function seekTranscript(currentTime) {
 
 /** Bolds the text in `transcriptLineLiElement` */
```
billie-cat, line 129: Update this comment to mention: Bolds the text if not already bold.

amuamushu: done, thanks for catching that!

PR #202, opened 2020-08-13
Update index.yaml to new Lecture and Comment field names

n-wach:
We need to update our indexes to match the updated names in Lecture and Comment.


billie-cat:
Please update the PR title. It is not very informative.

PR #205, opened 2020-08-13
Update height of the transcript container.

amuamushu:
Currently, the transcript container is cut off on smaller screens. This PR updates the height to prevent that from happening.


amuamushu:
> nit: maybe have a shorter title and move that information to the description

Thanks for pointing that out! I updated the title and description. :)

Issue #196, opened 2020-08-13
Remove login/logout buttons


n-wach:
All users viewing the deployed site will already be required to log in.

PR #207, opened 2020-08-13
Remove logout/login button

n-wach:
Closes: #196


f-camp:
Do we want to keep the ability to logout? What if the user want's to logout from the page after watching the lecture? How would they do this?

n-wach:
@f-camp Google decided that only Googlers can view our website, because it collects user data.  To enforce this, anyone viewing the site must be logged into an @google.com account.  Nobody can be logged out and view the site, so there isn't really a point in including a button...

PR #212, opened 2020-08-13
Improve responsiveness of synchronization

n-wach:
We want to make the synchronization feel more smooth:

- We increased the frequency of sync to 10 times per second (this can be reduced).

- In order to be able to update the time when a user seeks the video, the timer needs to be always running.  We added a check to ensure we don't unnecessarily seek (if the time has changed since last seek).

Although the timer is always running, CPU usage remains at 0% when the video is paused.


n-wach:
This accidentally duplicated changes from several branches, closing and recreating.

PR #213, opened 2020-08-13
Added lecture name to the header of view page

jaysaleh:
This pr adds the lecture name to the header of the view page.

### DIFF
--- a/src/main/webapp/view/view.js
+++ b/src/main/webapp/view/view.js
@@ -30,6 +30,7 @@ getLecture().then((lecture) => {
  * and transcript sections for the lecture view page.
  */
 async function initialize() {
+  setLectureName();
   window.loadVideoApi();
   window.intializeDiscussion();
   window.loadTranscript(window.location.search);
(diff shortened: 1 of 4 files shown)


src/main/webapp/view/view.js
```diff
@@ -53,3 +54,9 @@ function getLectureId() {
   const urlParams = new URLSearchParams(window.location.search);
   return urlParams.get(PARAM_ID);
 }
+
+/** Sets the lecture name in `header-text`. */
+function setLectureName() {
+  const headerText = document.getElementById('header-text');
+  headerText.innerText = window.LECTURE.lectureName;
```
n-wach, line 61: Can you remove the `href` attribute so clicking doesn't take us to the About page?

n-wach: https://developer.mozilla.org/en-US/docs/Web/API/Element/removeAttribute

jaysaleh: It is just a `<h2>` right now

n-wach: Ah yes, approved.

n-wach:
Oops, sorry Jay I accidentally merged my branch before this was in master.  Should still be fine tho, #217 was approved.

PR #214, opened 2020-08-13
Refactor seekTranscript to use helper methods.

amuamushu:
Update the structure of `seekTranscript` to use helper methods. The implementation of these helper methods will added in later pull requests.

Edit: Calling the helper methods inside `seekTranscript` would make more sense after the helper methods have been implemented. Thus, this PR will revert to draft until the helpers have been approved.
Edit 2: This PR will close for now because the structure of seekTranscript may change as other pull requests are reviewed so it would be easier to structure this pull request later.

### DIFF
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -119,9 +119,9 @@ function seekTranscript(currentTime) {
     addBold(currentTranscriptLine);
     return;
   }
-  const nextTranscript = getNextTranscriptLine(currentTimeMs);
+  const nextTranscriptLine = getNextTranscriptLine(currentTimeMs);
   removeBold(currentTranscriptLine);
-  currentTranscriptLine = nextTranscript;
+  currentTranscriptLine = nextTranscriptLine;
   scrollToTopOfTranscript(currentTranscriptLine);
   addBold(currentTranscriptLine);
 }
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -121,7 +121,7 @@ function seekTranscript(currentTime) {
     }
     return;
   }
-  let nextTranscript = getNextTranscriptLine(currentTimeMs);
+  const nextTranscript = getNextTranscriptLine(currentTimeMs);
   removeBold(currentTranscriptLine);
   currentTranscriptLine = nextTranscript;
   scrollToTopOfTranscript(currentTranscriptLine);
@@ -159,6 +159,7 @@ function isBeforeTheFirstTranscriptLine(currentTimeMs) {
   return currentTimeMs < firstTranscriptLine.startTimestampMs;
 }
 
+// TODO: Update this function to actually search for the next line.
 function getNextTranscriptLine(currentTimeMs) {
   let nextTranscript = currentTranscriptLine.nextElementSibling;
   if (currentTimeMs < currentTranscriptLine.startTimestampMs) {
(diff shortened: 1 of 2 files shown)


src/main/webapp/view/transcript/transcript.js
```diff
@@ -112,18 +112,18 @@ function deleteTranscript() {
...
+  if (isBeforeTheFirstTranscriptLine(currentTimeMs)) {
+    resetToBeforeTranscript();
   }
   if (isWithinCurrentTimeRange(currentTimeMs)) {
     addBold(currentTranscriptLine);
     return;
   }
+  const nextTranscript = getNextTranscriptLine(currentTimeMs);
   removeBold(currentTranscriptLine);
-  currentTranscriptLine = currentTranscriptLine.nextElementSibling;
```
n-wach, line 123: Can you move this into `getNextTranscriptLine` so we don't lose the seeking functionality we already have?  With a TODO to improve/change ofc

amuamushu: Done, thanks!

src/main/webapp/view/transcript/transcript.js
```diff
@@ -163,6 +163,26 @@ function isWithinCurrentTimeRange(currentTimeMs) {
...
+}
+
+/**
+ * Returns the closest transcript line based on `currentTimeMs`.
+ */
+function getNextTranscriptLine(currentTimeMs) {
+  // TODO: Implement this function to get
+  // the next line based on search.
+  currentTranscriptLine = currentTranscriptLine.nextElementSibling;
+  console.log(currentTimeMs);
```
jaysaleh, line 182: Is this needed?

amuamushu: Sort of, I needed to use the parameter for the checks to pass but I'll remove it when I implement the function.

src/main/webapp/view/transcript/transcript.js
```diff
@@ -112,18 +112,18 @@ function deleteTranscript() {
...
-  if (currentTimeMs < currentTranscriptLine.startTimestampMs) {
-    return;
+  if (isBeforeTheFirstTranscriptLine(currentTimeMs)) {
+    resetToBeforeTranscript();
   }
   if (isWithinCurrentTimeRange(currentTimeMs)) {
     addBold(currentTranscriptLine);
     return;
   }
+  const nextTranscript = getNextTranscriptLine(currentTimeMs);
```
jaysaleh, line 122: nit: rename to `nextTranscriptLine`

src/main/webapp/view/transcript/transcript.js
jaysaleh: Makes sense thanks

src/main/webapp/view/transcript/transcript.js
```diff
@@ -171,3 +191,12 @@ function scrollToTopOfTranscript(transcriptLine) {
   const ulElementOffset = transcriptLine.parentElement.offsetTop;
   transcriptContainer.scrollTop = transcriptLine.offsetTop - ulElementOffset;
 }
+
+/**
+ * Resets the transcript container to what it looks like before it
+ * started scrolling.
+ */
+function resetToBeforeTranscript() {
```
jaysaleh, line 199: Is this used right now? It doesn't do anything so maybe add it when you need it

amuamushu: Not really, but I added a function call for this in `seekTranscript` where I would like to call it once it is implemented. I thought of adding a stub as like a TODO. I'll keep it here if that's okay.

src/main/webapp/view/transcript/transcript.js
amuamushu: done! thanks

src/main/webapp/view/transcript/transcript.js
jaysaleh: Ah okay

PR #218, opened 2020-08-13
Improve responsiveness of synchronization

n-wach:
We want to make the synchronization feel more smooth:

- We increased the frequency of sync to 10 times per second (this can be reduced).

- In order to be able to update the time when a user seeks the video, the timer needs to be always running. We added a check to ensure we don't unnecessarily seek (if the time has changed since last seek).

Although the timer is always running, CPU usage remains at 0% when the video is paused.

### DIFF
--- a/src/main/webapp/synchronizer.js
+++ b/src/main/webapp/synchronizer.js
@@ -15,7 +15,7 @@
 let videoSyncTimer;
 let lastTime;
 
-const TIME_INTERVAL = 100;
+const TIME_INTERVAL_MS = 100;
 
 /**
  * Starts timer which broadcasts current video time every
@@ -24,7 +24,7 @@ const TIME_INTERVAL = 100;
 function startVideoSyncTimer() {
   videoSyncTimer = window.setInterval(() => {
     sync(window.videoPlayer.getCurrentTime());
-  }, /* ms= */ TIME_INTERVAL);
+  }, /* ms= */ TIME_INTERVAL_MS);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/webapp/synchronizer.js
```diff
@@ -25,16 +27,15 @@ function startVideoSyncTimer() {
...
 
-/** Stops video sync timer. */
-function stopVideoSyncTimer() {
-  clearInterval(videoSyncTimer);
-}
-
 /**
  * Calls functions that seek transcript, and discussion to {@code currentTime}
  * (number of seconds since video started playing).
  */
```
amuamushu, line 33: can you update the JSDoc to reflect how the calls are not made if the time has not changed?

n-wach: Done, thanks!

src/main/webapp/synchronizer.js
```diff
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 let videoSyncTimer;
-const TIME_INTERVAL = 1000;
+let lastTime;
+
+const TIME_INTERVAL = 100;
```
billie-cat, line 18: can you rename it to TIME_INTERVAL_MS so we know this is milliseconds.

n-wach:
> Why aren't we stopping the timer when the video is paused any more?
@f-camp 

Unfortunately there isn't an event for when the user drags the little slider around, and we think it is important to sync in this case.  The only way is to constantly poll `.getCurrentTime()`.  But with the `last == current` check we don't do anything unless the time actually changed

src/main/webapp/synchronizer.js
n-wach: Done!

PR #220, opened 2020-08-13
Add function for scrolling the transcript to the very beginning.

amuamushu:
This pull request creates a function will be called in `seekTranscript` when the current video time is before the start of any transcript line. It will ensure that the transcript is scrolled to the very top and the current line is no longer highlighted.

Edit: Having a separate set of function calls for times before the transcript is unnecessary. Instead, it would be cleaner to deal when searching for the closest line. Thus, this pull request will close.

### DIFF
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -172,3 +172,15 @@ function scrollToTopOfTranscript(transcriptLine) {
   const ulElementOffset = transcriptLine.parentElement.offsetTop;
   transcriptContainer.scrollTop = transcriptLine.offsetTop - ulElementOffset;
 }
+
+/**
+ * Resets the transcript container to what it looks like before it
+ * started scrolling.
+ */
+function resetToBeforeTranscript() {
+  // TODO: Create a static reference to the first Transcript line once 
+  // a class is created for transcript line.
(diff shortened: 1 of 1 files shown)


src/main/webapp/view/transcript/transcript.js
```diff
@@ -172,3 +172,15 @@ function scrollToTopOfTranscript(transcriptLine) {
   const ulElementOffset = transcriptLine.parentElement.offsetTop;
   transcriptContainer.scrollTop = transcriptLine.offsetTop - ulElementOffset;
 }
+
+/**
+ * Resets the transcript container to what it looks like before it
```
jaysaleh, line 177: This comment is a little confusing can you please rephrase this

n-wach: Also PR name

src/main/webapp/view/transcript/transcript.js
```diff
@@ -172,3 +172,15 @@ function scrollToTopOfTranscript(transcriptLine) {
...
+ * started scrolling.
+ */
+function resetToBeforeTranscript() {
+  // TODO: Create a static reference to the first Transcript line once
+  // a class is created for transcript line.
```
n-wach, line 182: Can this TODO be addressed now by updating base branch + merge?

PR #223, opened 2020-08-14
Add a favicon for zoomtube.

amuamushu:
zoomtube_favicon


src/main/webapp/index.html
```diff
@@ -2,6 +2,7 @@
 <meta charset="utf-8" />
 
 <title>ZoomTube</title>
+<link rel="icon" type="image/png" href="/favicon.png" />
```
billie-cat, line 5: what does fav mean?

n-wach: It's short for "favorite icon":  https://en.wikipedia.org/wiki/Favicon

PR #228, opened 2020-08-14
Add function to create a time range string given the start and end time.

amuamushu:
This function will be used to create the time range for the transcript.


src/main/webapp/timestamps.js
```diff
@@ -39,3 +39,13 @@ function timestampToString(timestampMs) {
 function secondsToMilliseconds(seconds) {
   return Math.round(seconds * MILLISECONDS_PER_SECOND);
 }
+
+/**
+ * Returns {String} A time range from
+ * `startTimestampMs` to `endTimestampMs`.
+ */
+function createTimestampRange(startTimestampMs, endTimestampMs) {
```
n-wach, line 47: `timestampRangeToString`?

src/main/webapp/timestamps.js
```diff
@@ -39,3 +39,13 @@ function timestampToString(timestampMs) {
 function secondsToMilliseconds(seconds) {
   return Math.round(seconds * MILLISECONDS_PER_SECOND);
 }
+
+/**
+ * Returns {String} A time range from
```
n-wach, line 44: Is putting `{String}` here recommended by the style guide? never seen it before

amuamushu: I believe so: https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments

```
In methods and named functions, parameter and return types must be documented, 
except in the case of same-signature @overrides, where all types are omitted. 
```

src/main/webapp/timestamps.js
amuamushu: done, thanks!

src/main/webapp/timestamps.js
```diff
@@ -39,3 +39,13 @@ function timestampToString(timestampMs) {
 function secondsToMilliseconds(seconds) {
   return Math.round(seconds * MILLISECONDS_PER_SECOND);
 }
+
+/**
+ * Returns {string} A time range from
```
n-wach, line 44: Can you do `Returns a time range {string} from ...` to make it flow better? Or is this order required.

n-wach: Just read that part of the styleguide again and it looks like `@return {string} A time range string created from ...` would be more typical

amuamushu: The order I have is what I've seen in the [style guide](https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments) so I think it is better leave it as is.

src/main/webapp/timestamps.js
```diff
@@ -39,3 +39,13 @@ function timestampToString(timestampMs) {
 function secondsToMilliseconds(seconds) {
   return Math.round(seconds * MILLISECONDS_PER_SECOND);
 }
+
+/**
+ * @returns {string} A time range from
```
billie-cat, line 44: You do not need to use @return if you only have 1 line of comment. Please write it as:

Returns a time range string in the forma of "`startTimestampMs` - `endTimestampMs`"

Or something like that.

f-camp: +1

amuamushu: done, thanks!

PR #234, opened 2020-08-17
Makes .js files modules

jaysaleh:
This PR makes all `.js `files modules such that functions can be imported/exported to other `.js` files, eliminating the need for `window.`


.eslintrc.json
```diff
@@ -12,6 +13,6 @@
     "no-multi-spaces": "off",
     "require-jsdoc": "off",
     "valid-jsdoc": "off",
-    "no-unused-vars": ["error", { "vars": "local", "args": "after-used", "ignoreRestSiblings": false }]
+    "no-unused-vars": ["off", { "vars": "local", "args": "after-used", "ignoreRestSiblings": false }]
```
n-wach, line 16: Can this be switched back to error?

src/main/webapp/view/view.js
```diff
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {intializeDiscussion} from '../view/discussion/discussion.js';
+
```
n-wach, line 16: Can the newlines between imports be removed? Or was that the linter

jaysaleh: That was the linter

src/main/webapp/view/discussion/discussion.js
```diff
@@ -39,6 +39,8 @@ const SELECTOR_REPLY_TEXTAREA = '#reply-textarea';
 // 10 seconds.
 const TIME_TOLERANCE_MS = 10000;
 
+window.postNewComment = postNewComment;
```
f-camp, line 42: Should we be able to eliminate window. here?

n-wach: Unfortunately not at the moment, because it is used in the HTML

PR #237, opened 2020-08-17
Add babel-eslint as parser

n-wach:
We need to switch eslint parsers to be able to use class fields (private and static)


jaysaleh:
Please resolve conflicts. Master has some validation errors right now that are being fixed in #239

n-wach:
@jaysaleh Fixed by changing base branch.  Will have to wait until #239 is merged for this one.

PR #240, opened 2020-08-17
Capitalizes .js files

jaysaleh:
This pull request capitalizes the `.js` file names.


src/main/webapp/view/index.html
```diff
@@ -11,12 +11,12 @@
...
-<script type="module" src="/view/discussion/discussion.js" defer></script>
-<script type="module" src="/view/video/video.js" defer></script>
-<script type="module" src="/view/transcript/transcript.js" defer></script>
-<script type="module" src="/view/view.js" defer></script>
+<script type="module" src="/Synchronizer.js" defer></script>
+<script type="module" src="/Timestamps.js" defer></script>
+<script type="module" src="/view/discussion/Discussion.js" defer></script>
+<script type="module" src="/view/video/Video.js" defer></script>
+<script type="module" src="/view/transcript/Transcript.js" defer></script>
+<script type="module" src="/view/View.js" defer></script>
```
amuamushu, line 19: I think you're missing the script tag for `Lectures.js`. It should be in a different index.html

n-wach:
Yeah I actually don't think we want to do this...  The javascript styleguide only requires uppercase for class names.  It doesn't say anything for modules, and every module I can find online uses lowercase

jaysaleh:
NVM we dont need this

PR #251, opened 2020-08-18
Create DiscussionManager javascript class

n-wach:
This PR is the first for refactoring discussion.js into classes.

The DiscussionManager handles fetching and posting comments from the `/discussions` endpoint.

Note: Some uses of this class are a little wacky right now.  This should improve as more classes are created and we can reuse the same manager instance by storing it as a field.


src/main/webapp/view/discussion/discussion.js
```diff
@@ -200,20 +115,130 @@ function getNearbyDiscussionComments(timestampMs) {
   return nearby;
 }
 
+
+class DiscussionManager {
```
jaysaleh, line 119: Move to separate file

src/main/webapp/view/discussion/discussion.js
```diff
@@ -200,20 +115,130 @@ function getNearbyDiscussionComments(timestampMs) {
...
+      [DiscussionManager.#PARAM_TYPE]: COMMENT_TYPE_REPLY,
+    });
+  }
+}
+
+
 /**
  * Renders a comment and its replies, with a form to post a new reply.
  */
 class DiscussionComment extends HTMLElement {
```
jaysaleh, line 227: Same here, move to separate file

src/main/webapp/view/discussion/discussion.js
```diff
@@ -200,20 +115,130 @@ function getNearbyDiscussionComments(timestampMs) {
...
 }
 
+
+class DiscussionManager {
+  static #ENDPOINT = '/discussion';
+  static #PARAM_LECTURE = 'lecture';
+  static #PARAM_PARENT = 'parent';
+  static #PARAM_TIMESTAMP = 'timestamp';
+  static #PARAM_TYPE = 'type';
+  #lecture;
```
amuamushu, line 125: Just curious, what does the `#` do?

n-wach: Marks a field as private, so supposedly it can only be accessed within the class.

src/main/webapp/view/discussion/discussion.js
n-wach: Done

src/main/webapp/view/discussion/discussion.js
n-wach: Going to wait on moving this one for a future PR.

src/main/webapp/view/discussion/discussion.js
```diff
@@ -14,12 +14,11 @@
 
 import {secondsToMilliseconds, timestampToString} from '../../timestamps.js';
```
n-wach, line 16: This newline was added by linter

src/main/webapp/view/discussion/discussion.js
f-camp: Please add a TODO for this, in general if there is anything that will be done in the future for discussion.js a TODO will be most helpful

src/main/webapp/view/discussion/discussion.js
```diff
@@ -256,7 +175,10 @@ class DiscussionComment extends HTMLElement {
     };
     this.shadowRoot.querySelector(SELECTOR_POST_REPLY).onclick = () => {
       const textarea = this.shadowRoot.querySelector(SELECTOR_REPLY_TEXTAREA);
-      postReply(textarea, this.comment.commentKey.id);
+      this.#manager.postReply(textarea.value, this.comment.commentKey.id)
+          .then(() => {
+            loadDiscussion();
+          });
```
f-camp, line 181: Work in constructor? Is postReply communicating with the server some how?

src/main/webapp/view/discussion/discussion-manager.js
```diff
@@ -0,0 +1,126 @@
...
+        rootComments.push(comment);
+      }
+    }
+    // Sort comments such that earliest timestamp is first.
+    rootComments.sort((a, b) => (a.timestampMs.value - b.timestampMs.value));
+    return rootComments;
+  }
+
+  /**
+   * Fetches and returns all comments in `this.#lecture` from the `ENDPOINT`.
```
billie-cat, line 65: Methods should never reference internal variables. You can make this doc simpler:

Fetches all of the lectures comments from the `ENDPOINT` and returns them sorted by timestamp.

Or something like that.

src/main/webapp/view/discussion/discussion-manager.js
```diff
@@ -0,0 +1,126 @@
...
+        DiscussionManager.#PARAM_LECTURE, this.#lecture.key.id);
+
+    const request = await fetch(url);
+    const json = await request.json();
+
+    return this.structureComments(json);
+  }
+
+  /**
+   * Posts `content` reloads the discussion. Adds query
```
billie-cat, line 81: this first sentence is confusing. Should it be Post `content` and reloads the discussion?

src/main/webapp/view/discussion/discussion-manager.js
```diff
@@ -0,0 +1,126 @@
...
+   * parameters from `params` to the request. Different types of comments
+   * require different parameters, such as `PARAM_TIMESTAMP` or `PARAM_PARENT`.
+   * The caller should ensure the correct parameters are supplied for the type
+   * of comment being posted.
```
billie-cat, line 85: Different types of comments...

This is extra detail and can probably be separated by <p>

src/main/webapp/view/discussion/discussion-manager.js
```diff
@@ -0,0 +1,126 @@
...
+   */
+  async postComment(content, params) {
```
billie-cat, line 87: I know we cannot make this method private, but can we comment that it is private. Also The order in which we declare methods is important in indicating what is more important. I would put this after the postRootComment and postReply methods.

src/main/webapp/view/discussion/discussion.js
n-wach: This is still setting up the functionality of the shadow DOM.  It just registers the event listeners, doesn't actually call them.

src/main/webapp/view/discussion/discussion.js
n-wach: Added

src/main/webapp/view/discussion/discussion-manager.js
n-wach: Added!

src/main/webapp/view/discussion/discussion-manager.js
n-wach: Done

src/main/webapp/view/discussion/discussion-manager.js
n-wach: Oops, reworded.

src/main/webapp/view/discussion/discussion-manager.js
n-wach: Changed, thanks!

src/main/webapp/view/discussion/discussion.js
n-wach: This particular section says:
When the POST_REPLY button in the shadowRoot is pressed, post a reply then reload the discussion.

f-camp: It still seems like work as I look through what postReply() does, I see that postComment() does some form of for loop or so. Maybe we can have an initialize() function here

n-wach: We chatted and I made the suggested changes in #264 .  Added a TODO here

PR #252, opened 2020-08-18
Refactors synchronizer.js file to include a class

jaysaleh:
This pull request refactors `synchronizer.js`  to include a synchronizer object


src/main/webapp/synchronizer.js
```diff
@@ -19,26 +19,28 @@ let lastTime;
 
 const TIME_INTERVAL_MS = 100;
```
amuamushu, line 20: should this constant be inside the class?

src/main/webapp/view/video/video.js
```diff
@@ -45,7 +43,7 @@ function onYouTubeIframeAPIReady() {
 /** `event` plays the YouTube video. */
 function onPlayerReady(event) {
   event.target.playVideo();
-  startVideoSyncTimer();
+  window.synchronizer.startVideoSyncTimer();
```
f-camp, line 46: Why window here? Wht not pass this into the video.js class? If this is the plan please add a TODO.

jaysaleh: A TODO for this is located in `View.js` where `synchronizer` is created. Do I need to add a TODO everywhere I do `window.synchronizer`?

src/main/webapp/synchronizer.js
```diff
@@ -19,26 +19,28 @@ let lastTime;
...
-/**
- * Starts timer which broadcasts current video time every
- * `TIME_INTERVAL` milliseconds.
- */
-export function startVideoSyncTimer() {
-  window.setInterval(() => {
-    sync(window.videoPlayer.getCurrentTime());
-  }, /* ms= */ TIME_INTERVAL_MS);
-}
+export default class Synchronizer {
```
billie-cat, line 22: Remember that classes should have JsDocs

https://google.github.io/styleguide/jsguide.html#jsdoc-class-comments

src/main/webapp/synchronizer.js
```diff
@@ -19,26 +19,28 @@ let lastTime;
...
-/**
- * Calls functions that seek transcript, and discussion to `currentTime`
- * (number of seconds since start of video), when the `currentTime`
- * changes from the last time this was called.
- */
-function sync(currentTime) {
-  if (currentTime == lastTime) {
-    return;
+  /**
+   * Calls functions that seek transcript, and discussion to `currentTime`
```
billie-cat, line 34:
```suggestion
   * Calls functions that seek transcript and discussion to `currentTime`
```

src/main/webapp/synchronizer.js
```diff
@@ -19,26 +19,28 @@ let lastTime;
...
+   * (number of seconds since start of video), when the `currentTime`
```
billie-cat, line 35: current time is in seconds? Why is that?

src/main/webapp/view/view.js
```diff
@@ -11,6 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+// module 'view.js'
```
billie-cat, line 14: What is this comment for? If it is necessary I would add a new line between this comment and the copyright or else it gets lost with the copyright.

jaysaleh: Sorry this is an error

src/main/webapp/synchronizer.js
jaysaleh: The video player returns the current time in seconds from the start of the video, the transcript and the discussion areas take the seconds and convert to ms. We will be converting this to ms in the future.

Added a TODO and #267

jaysaleh: @n-wach is working on this right now

src/main/webapp/view/video/video.js
f-camp: Ah I see, as long as there is a TODO.

src/main/webapp/synchronizer.js
billie-cat: Todo sounds great!

src/main/webapp/synchronizer.js
```diff
@@ -19,26 +19,30 @@ let lastTime;
...
-/**
- * Starts timer which broadcasts current video time every
- * `TIME_INTERVAL` milliseconds.
- */
-export function startVideoSyncTimer() {
-  window.setInterval(() => {
-    sync(window.videoPlayer.getCurrentTime());
-  }, /* ms= */ TIME_INTERVAL_MS);
-}
+/** Handles when to seek transcript and discussion areas. */
```
billie-cat, line 22:
```suggestion
/** Handles when to seek transcript and discussion areas according to video time. */
```

PR #255, opened 2020-08-18
Refactor loading of the Transcript Area into a class.

amuamushu:
This pull request moves the methods `loadTranscript()` and `addMultipleTranscriptLinesToDom()` into the class `TranscriptArea`, which will be responsible for loading the transcript.

Once #192 is merged into master, some of the TODOs will be addressed and reviewers will be pinged for a re-review.


src/main/webapp/view/transcript/transcriptArea.js
```diff
@@ -0,0 +1,76 @@
...
+    // eventController as the parameter.
+  }
+
+  /**
+   * Fetches the transcript lines from `ENDPOINT_TRANSCRIPT`.
+   *
+   * <p>This function assumes that the transcript lines have already
+   * been added to the datastore.
+   */
+  loadTranscript() {
```
n-wach: Would it be possible to make this `async` (and get rid of the `.then`s) so we can await it?  Just saying because the discussion will have to know when transcript is loaded before it can add indicators and making this async would allow .then or await in the calling method

src/main/webapp/view/transcript/transcriptArea.js
```diff
@@ -0,0 +1,76 @@
...
+   * `loadTranscript()`.
+   */
+  static addMultipleTranscriptLinesToDom(transcriptLines) {
+    const transcriptContainer =
+        document.getElementById(TranscriptArea.#TRANSCRIPT_CONTAINER);
+    if (transcriptContainer.firstChild) {
+      transcriptContainer.removeChild(transcriptContainer.firstChild);
+    }
+    const ulElement = document.createElement('ul');
+    ulElement.class = 'mx-auto';
```
n-wach: Can this be included in the HTML, and you get/use the element without recreating it here?

src/main/webapp/view/transcript/transcript.js
```diff
@@ -15,49 +15,17 @@
...
-  });
-}
-
 /**
  * Creates an `<li>` element containing `transcriptLine`'s text, start
  * time, and end time and appends it to `ulElement`.
  */
-function appendTextToList(transcriptLine, ulElement) {
+// TODO: Delete this function once TranscriptLineElement is in master.
+export function appendTextToList(transcriptLine, ulElement) {
```
n-wach, line 28: Can this be moved into the other class?  Or is it used elsewhere

src/main/webapp/view/transcript/transcriptArea.js
```diff
@@ -0,0 +1,76 @@
...
+   * Adds `transcriptLines` to the DOM as list elements.
+   *
+   * <p>This is a private method that should only be called in
+   * `loadTranscript()`.
+   */
+  static addMultipleTranscriptLinesToDom(transcriptLines) {
+    const transcriptContainer =
+        document.getElementById(TranscriptArea.#TRANSCRIPT_CONTAINER);
+    if (transcriptContainer.firstChild) {
+      transcriptContainer.removeChild(transcriptContainer.firstChild);
```
n-wach: What's this `if` for?

src/main/webapp/view/transcript/transcript.js
amuamushu: I was planning on leaving it here for now because it will not be used after the pull request for `TranscriptLineElement` is merged into master. 

I didn't want to actually move it because it will be a lot of lines changed and I'm not planning on keeping it.

src/main/webapp/view/transcript/transcriptArea.js
amuamushu: I think so. I made an issue to deal wit this later (#257) if that's okay.

src/main/webapp/view/transcript/transcriptArea.js
amuamushu: Yeah, I believe so. I made an issue for it (#258) and I'll deal with it in another pull request if that's okay.

src/main/webapp/view/transcript/transcriptArea.js
amuamushu: I added a comment for it. 

It clears the transcript container so there won't be duplicate transcript lines every time the page is refreshed.

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -0,0 +1,79 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO: Update this import statement once TranscriptLineElement is in master.
+import {appendTextToList} from '/transcript.js';
+
+/** Loads the transcript lines onto the DOM. */
+export class TranscriptArea {
```
n-wach, line 19: Actually one more thing:  Please make this `export default class` because it's the only thing in this file.

src/main/webapp/view/view.js
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import {intializeDiscussion} from '../view/discussion/discussion.js';
-import {loadTranscript} from './transcript/transcript.js';
+import {TranscriptArea} from './transcript/transcript-area.js';
```
n-wach, line 16: With export default you can remove the brackets from this import.

src/main/webapp/view/transcript/transcript-area.js
amuamushu: gotcha, thanks for the info! Done!

src/main/webapp/view/view.js
amuamushu: gotcha!

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -0,0 +1,80 @@
...
+    // eventController as the parameter.
+  }
+
+  /**
+   * Fetches the transcript lines from `ENDPOINT_TRANSCRIPT`.
+   *
+   * <p>This function assumes that the transcript lines have already
+   * been added to the datastore.
+   */
+  loadTranscript() {
```
f-camp, line 46: I would recommend any fetches to the server be async

src/main/webapp/view/view.js
```diff
@@ -37,7 +37,10 @@ async function initialize() {
   setLectureName();
   loadVideoApi();
   intializeDiscussion();
-  loadTranscript();
+  // TODO: Move TranscriptArea initialization outside of initialize()
+  // and replace string parameter with a controller object.
+  const transcript = new TranscriptArea('event controller');
+  transcript.loadTranscript();
```
f-camp, line 43: Calling the server here, let this be async. As a standard we should not block the UI if a network call is being made.

src/main/webapp/view/transcript/transcript-area.js
amuamushu: I made an issue to deal with this in a later pull request if that is okay. (#257).

src/main/webapp/view/view.js
amuamushu: thanks for the info!

PR #263, opened 2020-08-19
Adds ability to get current video time

### LINKED ISSUE #259: Add a function that returns current video time in Video.js
Currently `Synchronizer.js` is using window to access the video player but with the addition of the `Video` class this is no longer possible

jaysaleh:
This pr fixes #259. Adds a function to the `Video` class that returns the current video time.


src/main/webapp/view/video/video.js
```diff
@@ -49,6 +47,11 @@ export default class Video {
     startVideoSyncTimer();
   }
 
+  /** Returns current video time of 'videoPlayer'. */
+  getCurrentVideoTime() {
```
n-wach, line 51: Can you include the units?

n-wach: getCurrentVideoTimeSeconds

src/main/webapp/view/video/video.js
```diff
@@ -50,6 +48,11 @@ export default class Video {
     startVideoSyncTimer();
   }
 
+  /** Returns current video time of 'videoPlayer'. */
+  getCurrentVideoTimeSeconds() {
```
billie-cat, line 52: Seems error prone to have some times be in ms and some be sec. Shouldn't everything be in ms?

jaysaleh: This is being fixed by @n-wach

PR #264, opened 2020-08-19
Move DiscussionComment class to own file

n-wach:
This also moves constants only used by this class, and makes them private fields.


src/main/webapp/view/discussion/discussion-comment.js
```diff
@@ -0,0 +1,146 @@
...
+    };
+    this.shadowRoot.querySelector(DiscussionComment.#SELECTOR_CANCEL_REPLY)
+        .onclick = () => {
+      $(replyForm).collapse('hide');
+    };
+    this.shadowRoot.querySelector(DiscussionComment.#SELECTOR_POST_REPLY)
+        .onclick = () => {
+      const textarea = this.shadowRoot.querySelector(
+          DiscussionComment.#SELECTOR_REPLY_TEXTAREA);
+      this.#discussion.postReply(textarea.value, this.comment.commentKey.id);
```
f-camp, line 103: Does this need to be in an ```initialize()``` function too?

n-wach: I created a new function for this event handler, lmk if it needs anything else

src/main/webapp/view/discussion/discussion-comment.js
```diff
@@ -0,0 +1,152 @@
...
+    }
+    return `${timestampPrefix}${username} on ${comment.created}`;
+  }
+
+  /**
+   * Adds event listeners and handlers for the reply button and form.
+   * This makes the form open when "Reply" is pressed, submit when
+   * "Post" is pressed, and close when "Cancel" is pressed.
+   */
+  addReplyEventListeners() {
```
billie-cat, line 86: Can this be marked static so it is clear in the constructor we aren't playing with the internal state?

billie-cat: You can pass in the shadowRoot.

n-wach: Talked offline, and this change would be too complicated--keeping as-is

Issue #210, opened 2020-08-13
Improve discussion seek scroll


n-wach:
The discussion seek works when the video plays normally, but when the user skips around, it doesn't synchronize very well.  We need to think about how synchronization should behave.

PR #265, opened 2020-08-19
Make discussion scroll-on-seek more intuitive

n-wach:
Now we simply scroll to the nearest comment.

Closes: #210

### DIFF
--- a/src/main/webapp/view/discussion/discussion-area.js
+++ b/src/main/webapp/view/discussion/discussion-area.js
@@ -24,7 +24,6 @@ export const ELEMENT_DISCUSSION =
 export default class DiscussionArea {
   static #ELEMENT_POST_TEXTAREA = document.querySelector('#post-textarea');
   static #ELEMENT_TIMESTAMP_SPAN = document.querySelector('#timestamp-span');
-  static #TIME_TOLERANCE_MS = 10000;  // 10 seconds.
   #lecture;
   #manager;
   #currentTimeMs;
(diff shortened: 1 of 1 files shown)


src/main/webapp/view/discussion/discussion-area.js
```diff
@@ -55,38 +54,51 @@ export default class DiscussionArea {
   }
 
   /**
-   * Returns an array of the `DiscussionComment`s with timestamps near
-   * `timestampMs`. This returns an empty array if no elements are nearby.
-   *
-   * <p>A comment is nearby if it is within `TIME_TOLERANCE_MS`.
+   * Returns an array of the `DiscussionComment`s with the nearest time to
+   * `timeMs`. This typically returns an array with a single element, but
+   * if there are multiple comments the same distance away, they will all be
```
amuamushu, line 59: Why do you need to keep track of all of the nearest comments? In `seek`, it only scrolls to the first comment.

n-wach: Correct.  For now, we only care about the single nearest.  

However, this PR also considers how we're gonna want to highlight comments in the future.  If there are 5 comments at the same timestamp, we want to highlight them all.

Scrolling uses a single one, highlighting will use multiple.

n-wach: Will add to PR description

src/main/webapp/view/discussion/discussion-area.js
```diff
@@ -55,38 +54,48 @@ export default class DiscussionArea {
...
```
billie-cat, line 59: nit: seems the second sentence is extra detail and can be added after <p>

PR #277, opened 2020-08-20
Add buttons for icon feedback

jaysaleh:
This pr adds buttons for the feedback icon segment of zoomtube.

Screenshots: https://screenshot.googleplex.com/87edb557-ec62-4fdb-bcc0-6c11743c2686.png


src/main/webapp/feedback/feedback.js
```diff
@@ -0,0 +1,28 @@
...
+
+window.iconOnClick = iconOnClick;
+
+/**
+ * Sends `iconType`, video time stamp of when icon was clicked,
+ * and lecture ID to be stored in database.
+ */
+function iconOnClick(iconType) {
+  const videoTimeStamp = window.videoPlayer.getCurrentTime();
+  const url = new URL('/icon-feedback', window.location.origin);
```
amuamushu, line 23: can you update the strings to be constants?

n-wach: Might be a good idea to bundle these constants and function in some sort of `Feedback` class

jaysaleh: I don't think a feedback class is needed right now, as this file expands and view gets refactored ill make this a class

src/main/webapp/view/index.html
```diff
@@ -29,7 +29,15 @@ <h2 id="header-text">ZoomTube</h2>
 
   <div id="page">
     <div class="halfpage" id="player-transcript">
-      <div id="player"></div>
+      <div id="video-icons">
+        <div id="player"></div>
+        <div id="icons">
+          <button type="button" onclick="iconOnClick('good')">Thumbs Up</button>
```
n-wach, line 35: I think these should be changed to match the capitalization of the `enum`.  `GOOD`, `BAD`, `TOO_FAST`, or whatever they are

jaysaleh: fixed

src/main/webapp/feedback/icon-feedback.js
```diff
@@ -0,0 +1,36 @@
...
+
+const ENDPOINT_FEEDBACK = '/icon-feedback';
+
+const PARAM_LECTURE_ID = 'lectureId';
+const PARAM_TIMESTAMP = 'timestamp';
+const PARAM_ICON_TYPE = 'iconType';
+
+export default class IconFeedback {
+  /**
+   * Sends `iconType`, video time stamp of when icon was clicked,
```
billie-cat, line 23:
```suggestion
   * Sends `iconType`, video timestamp of when icon was clicked,
```

PR #294, opened 2020-08-21
 Links Synchronizer to Video eliminating global scope

jaysaleh:
This pr links a `Synchronizer` object to a `Video` object eliminating global scope.


src/main/webapp/view/view.js
```diff
@@ -43,12 +35,16 @@ getLecture().then((lecture) => {
...
 async function initialize() {
   setLectureName();
-  window.video.loadVideoApi();
-  intializeDiscussion();
+
+  const video = new Video();
+  const transcript = new TranscriptArea('event controller');
+
+  await video.loadVideoApi();
   // TODO: Move TranscriptArea initialization outside of initialize()
```
amuamushu, line 43: nit: maybe keep the TODO closer to where `TranscriptArea` was created.

src/main/webapp/view/video/video.js
```diff
@@ -12,15 +12,22 @@
...
 // limitations under the License.
 
+import Synchronizer from '../../synchronizer.js';
 import {secondsToMilliseconds} from '../../timestamps.js';
 
 const SCRIPT = 'script';
 
 /** Initializes and stores video player information. */
 export default class Video {
+  /** Links a new `Synchronizer` to a `Video`. */
```
n-wach, line 22: Does this JSDoc need to be updated?

src/main/webapp/view/video/video.js
```diff
@@ -12,15 +12,22 @@
...
+  constructor() {
+    this.synchronizer = new Synchronizer(this);
```
n-wach, line 24: can this.synchronizer be private?

src/main/webapp/view/video/video.js
jaysaleh: rephrased

src/main/webapp/view/video/video.js
jaysaleh: yes fixed

src/main/webapp/synchronizer.js
```diff
@@ -29,13 +29,18 @@ const transcriptSeeker = new TranscriptSeeker('event controller');
  * time.
  */
 export default class Synchronizer {
+  /** Links a `Video` to a `Synchronizer`. */
```
billie-cat, line 32: nit: this constructor comment isn't necessary. I would omit.

src/main/webapp/view/video/video.js
```diff
@@ -12,15 +12,24 @@
...
 
 const SCRIPT = 'script';
 
 /** Initializes and stores video player information. */
 export default class Video {
+  #synchronizer;
+
+  /** Creates a new `Synchronizer` linked to `this`. */
+  constructor() {
+    this.#synchronizer = new Synchronizer(this);
```
billie-cat, line 26: Typically it is bad design to have a circular dependency. Either the synchronizer should have a reference to Video or video have a reference to synchronizer. Can you refactor the code to remove this circular dependency.

jaysaleh: Modified to only have video have a reference to synchronizer

billie-cat: Why are you still passing 'this' to the Synchronizer. Can that be removed?

jaysaleh: Removed

src/main/webapp/view/video/video.js
```diff
@@ -38,15 +46,16 @@ export default class Video {
       width: '640',
       videoId: window.LECTURE.videoId,
       events: {
-        onReady: this.onPlayerReady,
+        onReady: window.onPlayerReady,
```
billie-cat, line 49: which is this window now? the method is declared in this class.

jaysaleh: The window is needed for the API to be able to call this

billie-cat: talked offline and Jay explained why this is needed.

PR #302, opened 2020-08-24
Display an error message if there is no transcript available for the lecture.

amuamushu:
Error message in the transcript container; https://screenshot.googleplex.com/dadcf029-719c-44b2-afa1-0c56d9637cfb.png

Additionally, a check is added in seekTranscript to ensure that the method body is only executed if there are transcript lines on the DOM. This prevents errors from occuring as `seekTranscript` is called.

This pull request bases off of transcript-loadTranscript which will be merged into master once pull request #275 is approved.

Edit: This code may have to be refactored when the seeking code is added.


src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -20,10 +20,14 @@ export default class TranscriptArea {
   static #ENDPOINT_TRANSCRIPT = '/transcript';
   static #TRANSCRIPT_CONTAINER = 'transcript-lines-container';
   static #PARAM_ID = 'id';
+  static #TRANSCRIPT_ERROR_MESSAGE =
+      'Sorry, there is no transcript available for this lecture recording. :(';
 
+  static #hasTranscript = true;
```
jaysaleh, line 26: Is it possible for this to be set when you check if a transcript exists or not? If we start seeking before we know a transcript actually exists will there be an issue?

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -49,9 +53,27 @@ export default class TranscriptArea {
     url.searchParams.append(TranscriptArea.#PARAM_ID, window.LECTURE_ID);
     const transcriptResponse = await fetch(url);
     const transcriptLines = await transcriptResponse.json();
+    if (transcriptLines.length == 0) {
```
jaysaleh, line 56: Can you add a comment mentioning the error message case?

src/main/webapp/view/transcript/transcript-area.js
amuamushu: I believe so. I'll update it to be set there when #286 is in master (PR that creates the scroll container). I added a TODO. There will be an error message displayed if it's seeked beforehand.

jaysaleh: Sounds good

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -38,7 +43,9 @@ export default class TranscriptArea {
   }
 
   /**
-   * Fetches the transcript lines from `ENDPOINT_TRANSCRIPT`.
+   * Fetches the transcript lines from `ENDPOINT_TRANSCRIPT`. If there
+   * are no transcript lines, an error message is displayed in the
+   * transcript container instead.
    *
    * <p>This function assumes that the transcript lines have already
```
billie-cat, line 50: is this still true?

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -49,9 +56,28 @@ export default class TranscriptArea {
     url.searchParams.append(TranscriptArea.#PARAM_ID, window.LECTURE_ID);
     const transcriptResponse = await fetch(url);
     const transcriptLines = await transcriptResponse.json();
+    // No transcript lines are available for this lecture.
+    if (transcriptLines.length == 0) {
```
billie-cat, line 60: Can't you set hasTranscript here? instead of making it static and true?

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -49,9 +56,28 @@ export default class TranscriptArea {
...
+   * transcript available for the lecture recording.
+   */
+  static displayNoTranscriptMessage() {
+    // TODO: Get the transcript container from a getter method once #286
+    // is merged.
+    const transcriptContainer =
+        document.getElementById(TranscriptArea.#TRANSCRIPT_CONTAINER);
+    transcriptContainer.innerText = TranscriptArea.#TRANSCRIPT_ERROR_MESSAGE;
+    transcriptContainer.classList.add('text-center');
+    TranscriptArea.#hasTranscript = false;
```
billie-cat, line 78: This has nothing to do with displaying there is no transcript message. Seems like a scope creep.

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -20,10 +20,15 @@ export default class TranscriptArea {
   static #ENDPOINT_TRANSCRIPT = '/transcript';
   static #TRANSCRIPT_CONTAINER = 'transcript-lines-container';
   static #PARAM_ID = 'id';
-
+  static #TRANSCRIPT_ERROR_MESSAGE =
+      'Sorry, there is no transcript available for this lecture recording. :(';
+  // TODO: Update hasTranscript in the getter method for the transcript
+  // container once #286 is merged.
```
billie-cat, line 26: Is this TODO still true?

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -20,10 +20,15 @@ export default class TranscriptArea {
...
+  static #hasTranscript = true;
```
billie-cat, line 27: Why is this field static? It is being set internally so I am not sure why it is static.

src/main/webapp/view/transcript/transcript-area.js
amuamushu: I ended up getting rid of the variable and checking if `currentTranscriptLine()` returns null instead because that should also handle the case where the transcript has not finished loading.

PR #306, opened 2020-08-25
Add comment indicator to trancriptLineElement.

amuamushu:
This adds the transcript indicator icon in each transcriptLineElement. By default, the indicators are invisible. In the future, the indicators will be visible when its comment amount is greater than 0.

https://screenshot.googleplex.com/7eBLEcVcNwAiGzw.png

### DIFF
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -100,7 +100,7 @@ export class TranscriptLineElement extends HTMLElement {
     const commentIndicator = document.createElement('span');
     commentIndicator.className =
         TranscriptLineElement.#COMMENT_INDICATOR_CLASSES;
-    commentIndicator.innerText = 12;
+    commentIndicator.innerText = 0;
     return commentIndicator;
   }
 
--- a/src/main/webapp/view/transcript/transcript.js
+++ b/src/main/webapp/view/transcript/transcript.js
@@ -33,6 +33,8 @@ export class TranscriptLineElement extends HTMLElement {
   static #COMMENT_INDICATOR_CLASSES =
       'indicator badge badge-pill rounded-circle';
 
+  commentIndicator;
+
   /**
    * Creates a custom HTML element representing `transcriptLine` with
    * the text and time range appended to the element.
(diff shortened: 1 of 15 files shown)


src/main/webapp/view/transcript/transcript.js
```diff
@@ -59,6 +61,8 @@ export class TranscriptLineElement extends HTMLElement {
         ['justify-content-start', 'mb-1', 'transcript-line-timestamp']);
     TranscriptLineElement.appendParagraphToContainer(
         transcriptLine.content, contentDivElement, ['ml-4', 'mb-1']);
+    contentDivElement.appendChild(
```
n-wach, line 64: I assume you'll need to save a reference to this for incrementing later?

src/main/webapp/view/transcript/transcript.js
```diff
@@ -87,6 +91,15 @@ export class TranscriptLineElement extends HTMLElement {
     return hrElement;
   }
 
+  /**
+   * Creates a stylized comment indicator element.
+   */
+  static createCommentIndicator() {
+    const span = document.createElement('span');
```
jaysaleh, line 98: Can you rename this to be more specific?

amuamushu:
> Code LGTM, not sure how i feel about ellipses around the numbers. What does it look like if you remove `rounded-circle` class?

Here is a screenshot of that
https://screenshot.googleplex.com/UtBtFzLtE3bidHa.png
Top 3 bubbles is without rounded-circle and bottom 3 are with rounded-circle

src/main/webapp/view/transcript/transcript.js
amuamushu: I was planning on querying for it, but saving a reference would be faster. Thanks! updated.

src/main/webapp/view/transcript/transcript.js
amuamushu: Renamed to be `commentIndicator`, thanks!

src/main/webapp/view/transcript/transcript.js
```diff
@@ -87,6 +93,17 @@ export class TranscriptLineElement extends HTMLElement {
     return hrElement;
   }
 
+  /**
+   * Creates a stylized comment indicator element.
```
billie-cat, line 97: I was thinking the comment indicator could start out invisible so that is doesn't appear empty until the comments are counted. WDYT?

amuamushu: Ah, the comment indicators are initially invisible. I set the visibility in the css.
https://github.com/googleinterns/zoomtube/pull/306/files#diff-52940dbcf82096e65b2848e5b8b28e41R128
Should I mention that they are invisible in the method header?

billie-cat: I would expand on this method.

/**
 * Creates an invisible stylized comment indicator element.
 *
 * <p> Once the element is populated then it becomes visible.
 */

amuamushu: gotcha, I update it to be that. Thanks!

PR #282, opened 2020-08-21
Fix minor layout and formatting inconsistencies

n-wach:
This PR made a lot of minor changes to fix some layout bugs and make styling more consistent.  Notable changes:
- Centered Video
- Added bg-light box around Discussion Area
- Made padding around Transcript and Discussion the same
- Made borders around textareas consistent
- Aligned transcript line content

I could have tried to make these separate PRs, but they were heavily reliant on each other, and style changes don't typically need much review.

Screenshots: 
- Large screen: https://screenshot.googleplex.com/9214ffb7-6620-418f-8b90-f7827f8c1674
- Smaller screen: https://screenshot.googleplex.com/aa6c2839-cb43-4613-aea1-179e90130b00

Next steps:
- Introduce class for rounded buttons (stop overriding bootstrap's styles)
- Change buttons in Discussion Area to correct colors/styles (like btn-link for reply)

PR #309, opened 2020-08-25
Remove global button style overrides, and restyle discussion buttons

n-wach:
Continuation of restyling as started in #282 

Previously, `btn` style was being overrided globally, which made it impossible to use some built in button styles, like `btn-link`.  This PR introduces `zt-btn` for those custom styles. 

This also move some other global styles to their own page, which were messing with other things.  For instance, the lecture-list page can now scroll.

Screenshots
- Large screen: https://screenshot.googleplex.com/7XXqrmaP4KBtWp2
- Small screen: https://screenshot.googleplex.com/4oLUSfR6ifNDk5Y
- Close of up reply form (disregard reply form in other screenshots, they are outdated): https://screenshot.googleplex.com/BfQqZCGB7XdrHF3

This should be it for major restyling.

### DIFF
--- a/src/main/webapp/view/lecture-view.css
+++ b/src/main/webapp/view/lecture-view.css
@@ -1,5 +1,18 @@
 /* Layout */
 
+body,
+html {
+  height: 100vh;
+  max-height: 100vh;
+  max-width: 100vw;
+  width: 100vw;
+  overflow: hidden;
+}
+
+body {
+  margin-bottom: 60px !important;
+}
+
 .halfpage {
   height: calc(100vh - 56px);
   overflow: hidden;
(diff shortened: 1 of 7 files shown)


src/main/webapp/view/lecture-view.css
```diff
@@ -1,5 +1,18 @@
 /* Layout */
 
+body,
+html {
+  height: 100vh;
+  max-height: 100vh;
+  max-width: 100vw;
+  width: 100vw;
+  overflow: hidden;
```
amuamushu, line 9: don't forget to alphabetize the declarations :)

n-wach: i knew i'd miss one! thanks for catching it

PR #310, opened 2020-08-25
Add type buttons for posting a new comment

n-wach:
This also includes the type in the new comment post request.

Screenshot: https://screenshot.googleplex.com/6s2eK762vxxkvxh

Unfortunately the buttons are hard coded into the HTML.  We can add these programmatically in the future, if we need to.

### DIFF
--- a/src/main/webapp/view/discussion/discussion-area.js
+++ b/src/main/webapp/view/discussion/discussion-area.js
@@ -28,6 +28,10 @@ export default class DiscussionArea {
   static #ELEMENT_TIMESTAMP_SPAN = document.querySelector('#timestamp-span');
   static #ELEMENT_NEW_COMMENT_TYPES =
       document.querySelector('#new-comment-types');
+  /**
+   * A selector to query on `ELEMENT_NEW_COMMENT_TYPES`. It returns the
+   * selected type button in the new comment area.
+   */
   static #SELECTOR_SELECTED_TYPE = 'label.active > input';
 
   #lecture;
--- a/src/main/webapp/view/discussion/discussion-area.js
+++ b/src/main/webapp/view/discussion/discussion-area.js
@@ -15,7 +15,6 @@
 import {timestampToString} from '../../timestamps.js';
 import DiscussionComment from './discussion-comment.js';
 import DiscussionManager from './discussion-manager.js';
-import {COMMENT_TYPE_QUESTION} from './discussion.js';
 
 export const ELEMENT_DISCUSSION =
     document.querySelector('#discussion-comments');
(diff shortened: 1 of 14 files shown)


src/main/webapp/view/index.html
```diff
@@ -48,9 +48,19 @@ <h2 id="header-text">ZoomTube</h2>
...
-        <button class="zt-btn btn btn-primary mb-2 float-right" type="button" onclick="postNewComment()">
-          Post Comment
-        </button>
+        <div class="btn-group-toggle" data-toggle="buttons" id="new-comment-types">
+          <label class="badge btn btn-outline-danger btn-sm float-left"
+            ><input type="radio" value="QUESTION" />Question</label
+          >
+          <label class="badge btn btn-outline-secondary btn-sm float-left active"
+            ><input type="radio" value="NOTE" />Note</label
+          >
```
n-wach, line 57: The linter keeps making it ugly here... not sure if I can do anything about it

src/main/webapp/view/discussion/discussion-area.js
```diff
@@ -27,6 +26,10 @@ export const ELEMENT_DISCUSSION =
 export default class DiscussionArea {
   static #ELEMENT_POST_TEXTAREA = document.querySelector('#post-textarea');
   static #ELEMENT_TIMESTAMP_SPAN = document.querySelector('#timestamp-span');
+  static #ELEMENT_NEW_COMMENT_TYPES =
+      document.querySelector('#new-comment-types');
+  static #SELECTOR_SELECTED_TYPE = 'label.active > input';
```
jaysaleh, line 31: This is confusing can you add a comment here

src/main/webapp/view/discussion/discussion-area.js
```diff
@@ -138,10 +141,16 @@ export default class DiscussionArea {
    * Posts the comment in the new comment area, and reloads the discussion.
    */
   postNewComment() {
+    /* eslint-disable indent */
+    const type = DiscussionArea.#ELEMENT_NEW_COMMENT_TYPES
```
jaysaleh, line 145: type is generic please make it more specific

n-wach: renamed to `commentType`

src/main/webapp/view/discussion/discussion-area.js
n-wach: added one

src/main/webapp/view/index.html
billie-cat: if you add a new line between type and value does that fix things?

n-wach: yeah that helps.  I added some other newlines and now it looks good

Issue #284, opened 2020-08-21
Direct users to an error page if something went wrong with loading the lecture.


amuamushu:
Currently, users see the stack trace when there is an issue with loading the page. We would like to update the redirect to be a nicer looking page to make the user experience better.

amuamushu:
Since lectures are still created even with the transcript error, I believe we can just catch the error and display a message on the transcript container (which will be done in #285) instead of directing them to a whole different page.

PR #315, opened 2020-08-26
Send redirect even on transcript error

n-wach:
Now that we have a nice message for empty transcripts, it's ok to simply redirect the user to the lecture view when we fail to initialize the transcript.

Closes: #284

### DIFF
--- a/src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
+++ b/src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
@@ -94,7 +94,13 @@ public class LectureServlet extends HttpServlet {
     String lectureName = request.getParameter(PARAM_NAME);
     Entity lectureEntity = LectureUtil.createEntity(lectureName, videoUrl, videoId.get());
     datastore.put(lectureEntity);
-    initializeTranscript(lectureEntity);
+    try {
+      initializeTranscript(lectureEntity);
+    } catch (IOException | ServletException e) {
+      // If there was an error initializing the transcript, then this lecture won't have one.
+      // Luckily that's still ok, so we suppress these errors so we can redirect.
+      // We might want to log these somewhere, but that's beyond the scope of this project.
+    }
     response.sendRedirect(buildRedirectUrl(lectureEntity));
   }
 
--- a/src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
+++ b/src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
@@ -125,8 +125,7 @@ public final class LectureServletTest {
   }
 
   @Test
-  public void doPost_videoHasNoTranscript_shouldReturnLectureAnyways()
-      throws IOException, ServletException {
+  public void doPost_videoHasNoTranscript_shouldReturnLectureAnyways() throws Exception {
     when(request.getParameter(LectureServlet.PARAM_NAME)).thenReturn(TEST_NAME);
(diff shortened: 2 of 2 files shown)


src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -94,7 +94,13 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     String lectureName = request.getParameter(PARAM_NAME);
     Entity lectureEntity = LectureUtil.createEntity(lectureName, videoUrl, videoId.get());
     datastore.put(lectureEntity);
-    initializeTranscript(lectureEntity);
+    try {
```
amuamushu, line 97: Can you add a unit test for sending a video without a transcript?

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -94,7 +94,13 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
...
+      initializeTranscript(lectureEntity);
+    } catch (IOException | ServletException e) {
+      // If there was an error initializing the transcript, then this lecture won't have one.
```
amuamushu, line 100: Can you add a unit test for sending a video without a transcript?

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -94,7 +94,13 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     String lectureName = request.getParameter(PARAM_NAME);
     Entity lectureEntity = LectureUtil.createEntity(lectureName, videoUrl, videoId.get());
     datastore.put(lectureEntity);
-    initializeTranscript(lectureEntity);
+    try {
+      initializeTranscript(lectureEntity);
+    } catch (IOException | ServletException e) {
```
jaysaleh, line 99: nit: `e` can be more descriptive in case it is used in the future.

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
jaysaleh: ^^

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
n-wach: nah, i think it can be changed if it's ever needed, but for now i don't want to draw attention to it

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
n-wach: Added!

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -122,6 +124,19 @@ public void doPost_urlAlreadyInDatabase_shouldReturnLecture()
     verify(response).sendRedirect("/view/?id=1");
   }
 
+  @Test
+  public void doPost_videoHasNoTranscript_shouldReturnLectureAnyways()
+      throws IOException, ServletException {
```
f-camp, line 129: for tests do throws Exception.

n-wach: fixed!  I see a lot of other tests should make this change, so I created #355 to track

Issue #172, opened 2020-08-11
Remove line breaks from transcript lines


n-wach:
Remove the `<br>` elements--if lines get too long we can style them ourselves

PR #317, opened 2020-08-26
Remove newlines from transcript line content

n-wach:
Creates helper method in `TranscriptParser` that unescapes XML and removes newlines.  

Any additional transcript line processing can be added to this method.

Closes: #172


src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
```diff
@@ -164,4 +164,17 @@ private Entity createTranscriptLineFromElement(Key lectureKey, Element transcrip
...
+
+  /**
+   * Cleans the {@code content} of a transcript line by unescaping XML characters, and
+   * removing newlines.
+   */
+  private String cleanupTranscriptContent(String content) {
+    String unescapedXml = StringEscapeUtils.unescapeXml(content);
+
+    // We ignore '\r' for now.
+    String cleanedContent = unescapedXml.replace('\n', ' ');
```
jaysaleh, line 176: Is there a reason we replace with a space and not an ''?

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
```diff
@@ -164,4 +164,17 @@ private Entity createTranscriptLineFromElement(Key lectureKey, Element transcrip
     return TranscriptLineUtil.createEntity(
         lectureKey, lineContent, lineStartMs, lineDurationMs, lineEndMs);
   }
+
+  /**
+   * Cleans the {@code content} of a transcript line by unescaping XML characters, and
+   * removing newlines.
+   */
+  private String cleanupTranscriptContent(String content) {
+    String unescapedXml = StringEscapeUtils.unescapeXml(content);
```
jaysaleh, line 173: can you add transcript line to the name of this

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
```diff
@@ -123,22 +125,56 @@ private Document fetchUrlAsXmlDocument(URL url) throws IOException {
...
+
+    // We ignore '\r' for now.
+    String cleanedContent = unescapedXml.replace('\n', ' ');
+
+    return cleanedContent;
+  }
```
amuamushu, line 179: can a unit test be added to ensure that it is working?

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
n-wach: Yeah.  Example transcript line: `This is a super long transcript line\nthat has a newline.`

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
n-wach: not sure what you meant by this... I changed the names a little tho, tell me how it looks now

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
n-wach: I can using `@VisibleForTesting`, is that a good idea?

amuamushu: Ah I guess since it's a bit tricky to test, not having a unit test should be okay. We probably shouldn't test it directly with @visibleForTesting since private methods don't have to be tested.

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
jaysaleh: `unescapedTranscriptLine`

n-wach: Ah, will do

n-wach: I changed it to `unescapedContent` b/c it's the content, not the line as a whole.  do you think that's ok?

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
```diff
@@ -164,4 +164,17 @@ private Entity createTranscriptLineFromElement(Key lectureKey, Element transcrip
     return TranscriptLineUtil.createEntity(
         lectureKey, lineContent, lineStartMs, lineDurationMs, lineEndMs);
   }
+
+  /**
+   * Cleans the {@code content} of a transcript line by unescaping XML characters, and
```
billie-cat, line 169:
```suggestion
   * Cleans the {@code content} of a transcript line by unescaping XML characters and
```

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
billie-cat: Can't you load a transcript which has \n and then you obtain it and see there are no newlines? 

If you cannot easily test this then its a code smell that the class is doing too much.

billie-cat: @amuamushu  private methods should be tested, but through the public methods which call them.

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
jaysaleh: sure that'll work

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
n-wach: Done, although it's rather inelegant.  The only way to test this is with a video id, that results in an API call.  And the only video I know with newlines has a lot of transcript lines, so it increased the total test running time by half a second or so.

src/main/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParser.java
n-wach: Done!

PR #321, opened 2020-08-26
Add function for incrementing the comment indicator.

amuamushu:
This function will be called when iterating the new comments to increment the comment indicator.


src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -21,9 +21,12 @@ export default class TranscriptArea {
   static #TRANSCRIPT_CONTAINER = 'transcript-lines-container';
   static #PARAM_ID = 'id';
 
+  static transcriptLineMap = new Map();
```
jaysaleh, line 24: Is there a reason this isn't private?

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -72,15 +75,32 @@ export default class TranscriptArea {
...
+  static incrementCommentIndicatorAt(transcriptLineKeyId) {
+    if (!TranscriptArea.transcriptLineMap.has(transcriptLineKeyId)) {
+      return;
+    }
+    const commentIndicatorElement =
+        TranscriptArea.transcriptLineMap.get(transcriptLineKeyId)
+            .commentIndicator;
+    commentIndicatorElement.innerText =
+        parseInt(commentIndicatorElement.innerText) + 1;
+    commentIndicatorElement.style.visibility = 'visible';
```
jaysaleh, line 96: nit: can this be a const?

amuamushu: I think it's clearer what is happening to leave it as is if that's okay.

src/main/webapp/view/transcript/transcript-area.js
amuamushu: Ah, there isn't. I updated it to be private, thanks!

src/main/webapp/view/lecture-view.css
```diff
@@ -115,6 +115,19 @@ iframe {
   width: 100%;
 }
 
+/* Comment Indicator */
```
billie-cat, line 118: I think I already saw this code. Does it need to be synced?

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -21,9 +21,12 @@ export default class TranscriptArea {
   static #TRANSCRIPT_CONTAINER = 'transcript-lines-container';
   static #PARAM_ID = 'id';
 
+  static #transcriptLineMap = new Map();
```
billie-cat, line 24: Why is this static? 

Also, remember not to add redundant names like "Map". This is a map of transcriptLine to comment count maybe rename it transcriptLineToCommentCount

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -72,15 +75,32 @@ export default class TranscriptArea {
...
+      const transcriptLineElement =
+          TranscriptLineElement.createTranscriptLineElement(transcriptLine);
+      ulElement.appendChild(transcriptLineElement);
+      TranscriptArea.#transcriptLineMap.set(
+          transcriptLine.transcriptKey.id, transcriptLineElement);
     });
   }
 
+  /** Increments the indicator corresponding to `transcriptLineKey` by 1. */
+  static incrementCommentIndicatorAt(transcriptLineKeyId) {
```
billie-cat, line 87: transcriptLineMap shouldn't be static and so incrementCommentIndicatorAt shouldn't be static either.

billie-cat, line 87: where does this method get called?

amuamushu: gotcha, updated!

amuamushu: This method gets called in `DiscussionArea` as the new comments are inserted so I would have to pass a reference of `TranscriptArea` to `DiscussionArea`

src/main/webapp/view/lecture-view.css
amuamushu: sorry about that, just synced it. Thanks!

src/main/webapp/view/transcript/transcript-area.js
billie-cat: Why is the variable called transcriptLineKeyId and not transcriptLineKey? This does not match the JsDoc.

amuamushu: ahh, thanks for catching that!

I updated it to be `transcriptLineKeyId` because a transcriptLine key is formatted like this: {kind: "TranscriptLine", id: 123456}. It is easier to just hold onto the id since we know the kind is `transcriptLine`.

PR #322, opened 2020-08-26
Update current transcript to be static.

amuamushu:
This update makes the `currentTranscriptLine` more accessible outside of the current class. Since there is only one current transcript line for a video, it is okay to make `currentTranscriptLine` static.

Edit: This change will no longer be merged into master because we have decided it would be safer to pass a reference of `TranscriptSeeker` to the discussion area.

### DIFF
--- a/src/main/webapp/view/transcript/transcript-seeker.js
+++ b/src/main/webapp/view/transcript/transcript-seeker.js
@@ -81,7 +81,7 @@ export default class TranscriptSeeker {
     // is the one immediately after it. This check is done before
     // the search is conducted because it is more time efficient
     // to check the next element than to conduct a search.
-    if (nextTranscript && nextTranscript.isWithinTimeRange(timeMs)) {
+    if (nextTranscript != null && nextTranscript.isWithinTimeRange(timeMs)) {
       return nextTranscript;
     }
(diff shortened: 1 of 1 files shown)


f-camp:
It should be better to have a getter function. Making things static is unsafe as it makes them mutable outside of the class.

src/main/webapp/view/transcript/transcript-seeker.js
```diff
@@ -16,7 +16,7 @@
 export default class TranscriptSeeker {
   static #TRANSCRIPT_CONTAINER = 'transcript-lines-container';
 
-  #currentTranscriptLine;
+  static #currentTranscriptLine;
```
f-camp, line 19: Have a getter instead. Also isn't this static private? How would it be accessible outside the class? (I might be missing something)

billie-cat: I don't think we can make things private in js.

amuamushu: I think we just can't have private methods but we're allowed to have private variables. The # denotes that it is private.

amuamushu:
This change will no longer be merged into master because we have decided it would be safer to pass a reference of `TranscriptSeeker` to the discussion area.

PR #323, opened 2020-08-26
Add TranscriptLine field to Comment

n-wach:
This is used for saving where to add indicators.

The discussion currently has no way of getting the current transcript line, so there is a TODO.  Once there is a way to get the current transcript line, the new comment post request can be modified to use it, and this feature should be fully working.

### DIFF
--- a/src/main/java/com/googleinterns/zoomtube/data/Comment.java
+++ b/src/main/java/com/googleinterns/zoomtube/data/Comment.java
@@ -81,7 +81,6 @@ public abstract class Comment {
     public abstract Builder setParentKey(Optional<Key> parentKey);
     public abstract Builder setTimestampMs(long timestampMs);
     public abstract Builder setTimestampMs(Optional<Long> timestampMs);
-    public abstract Builder setTranscriptLineKey(Key transcriptLineKey);
     public abstract Builder setTranscriptLineKey(Optional<Key> transcriptLineKey);
     public abstract Builder setAuthor(User author);
     public abstract Builder setContent(String content);
--- a/src/main/java/com/googleinterns/zoomtube/data/Comment.java
+++ b/src/main/java/com/googleinterns/zoomtube/data/Comment.java
@@ -50,6 +50,11 @@ public abstract class Comment {
   public abstract Optional<Long> timestampMs();
 
   /**
+   * Returns the key of the transcript line where an indicator should appear.
+   */
+  public abstract Key transcriptLineKey();
+
+  /**
    * Returns the comment's author.
    */
   public abstract User author();
@@ -75,6 +80,7 @@ public abstract class Comment {
     public abstract Builder setParentKey(Optional<Key> parentKey);
     public abstract Builder setTimestampMs(long timestampMs);
     public abstract Builder setTimestampMs(Optional<Long> timestampMs);
+    public abstract Builder setTranscriptLineKey(Key transcriptLineKey);
     public abstract Builder setAuthor(User author);
     public abstract Builder setContent(String content);
     public abstract Builder setCreated(Date created);
--- a/src/main/java/com/googleinterns/zoomtube/data/Comment.java
+++ b/src/main/java/com/googleinterns/zoomtube/data/Comment.java
@@ -50,9 +50,10 @@ public abstract class Comment {
   public abstract Optional<Long> timestampMs();
 
   /**
-   * Returns the key of the transcript line where an indicator should appear.
+   * Returns the key of the comment's transcript line, or {@code Optional.empty()} if it
+   * doesn't have one.
    */
-  public abstract Key transcriptLineKey();
+  public abstract Optional<Key> transcriptLineKey();
 
   /**
    * Returns the comment's author.
@@ -81,6 +82,7 @@ public abstract class Comment {
     public abstract Builder setTimestampMs(long timestampMs);
     public abstract Builder setTimestampMs(Optional<Long> timestampMs);
     public abstract Builder setTranscriptLineKey(Key transcriptLineKey);
+    public abstract Builder setTranscriptLineKey(Optional<Key> transcriptLineKey);
     public abstract Builder setAuthor(User author);
     public abstract Builder setContent(String content);
     public abstract Builder setCreated(Date created);
--- a/src/main/java/com/googleinterns/zoomtube/data/Comment.java
+++ b/src/main/java/com/googleinterns/zoomtube/data/Comment.java
@@ -50,8 +50,8 @@ public abstract class Comment {
   public abstract Optional<Long> timestampMs();
 
   /**
-   * Returns the key of the comment's transcript line, or {@code Optional.empty()} if it
-   * doesn't have one.
+   * Returns the transcript line key for which this comment is for, or {@code Optional.empty()} if
+   * it doesn't have one.
    */
   public abstract Optional<Key> transcriptLineKey();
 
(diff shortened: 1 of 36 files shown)


src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -86,19 +89,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     final Entity commentEntity;
     long lectureId = Long.parseLong(request.getParameter(PARAM_LECTURE));
     Key lecture = KeyFactory.createKey(LectureUtil.KIND, lectureId);
-    // TODO: Require non-zero content length. See: #183.
+    // TODO: Validate that the transcript line contains the timestampMs.
+    long transcriptId = Long.parseLong(request.getParameter(PARAM_TRANSCRIPT_LINE));
+    Key transcriptLine = KeyFactory.createKey(TranscriptLineUtil.KIND, transcriptId);
```
jaysaleh, line 94: Can you name this `transcriptLineKey`?

jaysaleh, line 94: Just a note, in some places we have 'key' in the name of the key. Like in `CommentUtil.java` this would be named `transcriptLineKey`. Can we stay consistent to avoid confusing in the naming, can be done in a seperate pr.

n-wach: Good point, I'll make the change in this one.  Thanks!

src/main/java/com/googleinterns/zoomtube/data/Comment.java
```diff
@@ -49,6 +49,11 @@
    */
   public abstract Optional<Long> timestampMs();
 
+  /**
+   * Returns the key of the transcript line where an indicator should appear.
```
billie-cat, line 53: So for encapsulation the Comment knows nothing about the TranscriptLine indicator.  I would rephrase this as:

Returns the transcript line key for which this comment is for.

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -85,20 +88,23 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
 
     final Entity commentEntity;
     long lectureId = Long.parseLong(request.getParameter(PARAM_LECTURE));
-    Key lecture = KeyFactory.createKey(LectureUtil.KIND, lectureId);
-    // TODO: Require non-zero content length. See: #183.
+    Key lectureKey = KeyFactory.createKey(LectureUtil.KIND, lectureId);
+    // TODO: Validate that the transcript line contains the timestampMs.
+    long transcriptId = Long.parseLong(request.getParameter(PARAM_TRANSCRIPT_LINE));
+    Key transcriptLineKey = KeyFactory.createKey(TranscriptLineUtil.KIND, transcriptId);
```
billie-cat, line 94: If the lecture has no transcript, then does this still returns a transcriptId?

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -85,20 +88,23 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
 
     final Entity commentEntity;
     long lectureId = Long.parseLong(request.getParameter(PARAM_LECTURE));
-    Key lecture = KeyFactory.createKey(LectureUtil.KIND, lectureId);
-    // TODO: Require non-zero content length. See: #183.
+    Key lectureKey = KeyFactory.createKey(LectureUtil.KIND, lectureId);
+    // TODO: Validate that the transcript line contains the timestampMs.
+    long transcriptId = Long.parseLong(request.getParameter(PARAM_TRANSCRIPT_LINE));
```
billie-cat, line 93: I believe 'transcriptId' should be renamed 'transcriptLineId'. This is the id of the transcript line and not the transcript, right?

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
```diff
@@ -113,6 +119,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     if (request.getParameter(PARAM_TYPE) == null) {
       return Optional.of(ERROR_MISSING_COMMENT_TYPE);
     }
+    if (request.getParameter(PARAM_TRANSCRIPT_LINE) == null) {
+      return Optional.of(ERROR_MISSING_TRANSCRIPT_LINE);
+    }
```
billie-cat, line 124: Is this an error though? It is possible for a Lecture to have no transcript right?

src/main/webapp/view/discussion/discussion-manager.js
```diff
@@ -88,22 +89,24 @@ export default class DiscussionManager {
...
     await this.postComment(content, {
       [DiscussionManager.#PARAM_TIMESTAMP]: timestampMs,
       [DiscussionManager.#PARAM_TYPE]: type,
+      [DiscussionManager.#PARAM_TRANSCRIPT_LINE]: transcriptLineId,
     });
   }
 
   /**
-   * Posts `content` as a reply to `parentId`.
+   * Posts `content` as a reply to `parentId` with the specified `transcriptId`.
```
billie-cat, line 103: 'transcriptId' is a type.the field is transcriptLineId.

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: ah yes, fixed!

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: talked to amy about how we want to handle this, and had decided to make this param optional

src/main/java/com/googleinterns/zoomtube/data/Comment.java
n-wach: done!

src/main/java/com/googleinterns/zoomtube/servlets/DiscussionServlet.java
n-wach: talked to amy about how we want to handle this, and had decided to make this param optional

src/main/webapp/view/discussion/discussion-manager.js
n-wach: ah, whoops, fixed!

src/main/java/com/googleinterns/zoomtube/data/Comment.java
```diff
@@ -75,6 +81,8 @@ public static Builder builder() {
     public abstract Builder setParentKey(Optional<Key> parentKey);
     public abstract Builder setTimestampMs(long timestampMs);
     public abstract Builder setTimestampMs(Optional<Long> timestampMs);
+    public abstract Builder setTranscriptLineKey(Key transcriptLineKey);
+    public abstract Builder setTranscriptLineKey(Optional<Key> transcriptLineKey);
```
billie-cat, line 85: Why are these methods duplicated? You only need 1.

n-wach: I am having some compiler errors when I only include one of them...

n-wach: Removed the non-optional one after talking offline

Issue #185, opened 2020-08-12
Use a filter in checkUrlInDatabase to avoid fetching all lectures


n-wach:
We can add a filter to the videoUrl to look for the specific one that matches.

Also, consider looking for matching videoId instead, because that's what really matters

PR #326, opened 2020-08-26
Use Filter on videoId to detect duplicate lectures

n-wach:
This is cleaner and faster, and also checks the video id instead of URL (because there are so many possible URL forms for the same video).

Closes: #185

### DIFF
--- a/src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
+++ b/src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
@@ -22,6 +22,9 @@ import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
 import com.googleinterns.zoomtube.transcriptParser.TranscriptParser;
@@ -85,7 +88,7 @@ public class LectureServlet extends HttpServlet {
       return;
     }
 
-    Optional<Entity> existingEntity = checkUrlInDatabase(videoUrl);
+    Optional<Entity> existingEntity = checkVideoInDatabase(videoId.get());
     if (existingEntity.isPresent()) {
       response.sendRedirect(buildRedirectUrl(existingEntity.get()));
       return;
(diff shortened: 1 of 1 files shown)


src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -147,21 +150,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
 
   /**
-   * Returns the Entity in database that has {@code url}, or
+   * Returns the Entity in database that has {@code videoId}, or
    * {@code Optional.empty()} if one doesn't exist.
    */
-  // TODO: Use a filter to avoid fetching all lectures.  See: #185.
-  private Optional<Entity> checkUrlInDatabase(String url) {
-    Query query = new Query(LectureUtil.KIND);
+  private Optional<Entity> checkVideoInDatabase(String videoId) {
```
amuamushu, line 156: nit: I find it weird that the method name has the word "check" in it because at least to me, I feel that implies a boolean is returned. Would it be better to return a boolean instead of an optional considering `existingEntity` is only used for checking?

n-wach: Renamed to `queryForLectureWithVideoId`

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -147,21 +150,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
 
   /**
-   * Returns the Entity in database that has {@code url}, or
+   * Returns the Entity in database that has {@code videoId}, or
    * {@code Optional.empty()} if one doesn't exist.
    */
-  // TODO: Use a filter to avoid fetching all lectures.  See: #185.
-  private Optional<Entity> checkUrlInDatabase(String url) {
-    Query query = new Query(LectureUtil.KIND);
+  private Optional<Entity> queryForLectureWithVideoId(String videoId) {
```
billie-cat, line 156: Would a better method name be 'lectureWithVideoId'. "queryFor" seems to explain how you obtain the lecture, so I am not sure it is necessary.

n-wach: It doesn't always find/return a lecture with a matching video id.  The original name was `check...`, but we felt like that implied a boolean return value, and changed it to this.  If you think `lectureWithVideoId` is better i can change it. :)

billie-cat: Optional<Entity> implies it might not return a lecture with video id. This is a getter method and typically getter methods are called getFoo() or foo(). 

This is a nit mostly, so you can keep it as is if you prefer your name.

n-wach: This made more sense to us, so i'll keep it

Issue #307, opened 2020-08-25
Move reply form to below existing replies


n-wach:
New replies appear at the bottom, so it makes sense for the reply form to be there.

PR #327, opened 2020-08-27
Move comment reply form to end of replies

n-wach:
Because newer comments are now at the bottom, it makes sense for the reply form to appear where the comment will be.

Clicking "reply" opens the form at the bottom, and causes you to scroll and focus on the text area.

Closes: #307

### DIFF
--- a/src/main/webapp/view/discussion/discussion-area.js
+++ b/src/main/webapp/view/discussion/discussion-area.js
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 import {timestampToString} from '../../timestamps.js';
+import DiscussionComment from './discussion-comment.js';
 import DiscussionManager from './discussion-manager.js';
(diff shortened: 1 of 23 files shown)


src/main/webapp/view/discussion/discussion-area.js
```diff
@@ -40,32 +39,64 @@ export default class DiscussionArea {
...
-    this.#currentRootCommentElements = [];
     this.#nearestComments = [];
   }
 
   /**
    * Initialize the discussion area by loading the current comments.
    */
   async initialize() {
-    await this.loadDiscussion();
+    await this.updateDiscussion();
```
billie-cat, line 49: How is this related to your PR? Do you need to sync?

n-wach: Just merged, should only have the new changes now

PR #338, opened 2020-08-27
Allow users to seek by clicking on the transcript.

amuamushu:
With this change, when users click on the timestamp of a transcript line, the transcript jumps to that line. When the event handlers are added to the video and the discussion, they will also be seeked upon clicking the transcript timestamp.


src/main/webapp/scroll-container.js
```diff
@@ -72,7 +72,8 @@ export class ScrollContainer extends HTMLDivElement {
    * Otherwise, does nothing.
    *
    */
-  scrollToTopOfContainer(element, forceScroll = false) {
+  scrollToTopOfContainer(
+      element = this.lastChild.firstChild, forceScroll = false) {
```
jaysaleh, line 76: What is element here?

src/main/webapp/view/transcript/transcript-seeker.js
```diff
@@ -114,5 +117,7 @@ export default class TranscriptSeeker {
     return transcriptLinePointer;
   }
 
-  // TODO: Add a method that adds the eventListeners.
+  eventController() {
```
jaysaleh, line 120: Asking for context, what is this for?

src/main/webapp/scroll-container.js
n-wach: the `element` within the container that should be scrolled to... maybe `childElement` would be better?  but it isn't a direct child either

src/main/webapp/view/transcript/transcript.js
```diff
@@ -30,6 +30,9 @@ export function deleteTranscript() {
 export class TranscriptLineElement extends HTMLElement {
   static #DEFAULT_FONT_WEIGHT = 'text-muted';
   static #BOLD_FONT_WEIGHT = 'font-weight-bold';
+  static #TIME_OFFSET = 1;
+
+  #timestamp;
```
n-wach, line 35: can this be renamed `timestampElement`?  timestamp could be a lot of things

src/main/webapp/view/transcript/transcript.js
```diff
@@ -54,9 +57,14 @@ export class TranscriptLineElement extends HTMLElement {
...
     const contentDivElement = TranscriptLineElement.createContentDivElement();
-    TranscriptLineElement.appendParagraphToContainer(
-        timestampRange, contentDivElement,
-        ['justify-content-start', 'mb-1', 'transcript-line-timestamp']);
+    this.#timestamp =
+        TranscriptLineElement.createParagraphWithClasses(timestampRange, [
+          'justify-content-start',
+          'mb-1',
+          'transcript-line-timestamp',
+          'timestamp',
```
n-wach, line 65: what's the difference between transcript-line-timestamp and timestamp?  Can these CSS classes be merged?

src/main/webapp/view/transcript/transcript.js
```diff
@@ -94,14 +109,23 @@ export class TranscriptLineElement extends HTMLElement {
    * <p>Adds classes the the p tag if `classList` is provided.
    */
   static appendParagraphToContainer(text, container, classes = []) {
+    container.appendChild(
+        TranscriptLineElement.createParagraphWithClasses(text, classes));
+  }
+
+  /**
+   * Creates a p tag to store the given `text`.
```
n-wach, line 117: `p` tag

src/main/webapp/view/transcript/transcript.js
```diff
@@ -94,14 +109,23 @@ export class TranscriptLineElement extends HTMLElement {
...
+   *
+   * <p>Adds classes the the p tag if `classList` is provided.
```
n-wach, line 119: I assume this should be `to the`? and add ticks on `p` please

src/main/webapp/view/transcript/transcript.js
```diff
@@ -30,6 +30,9 @@ export function deleteTranscript() {
 export class TranscriptLineElement extends HTMLElement {
   static #DEFAULT_FONT_WEIGHT = 'text-muted';
   static #BOLD_FONT_WEIGHT = 'font-weight-bold';
+  static #TIME_OFFSET = 1;
```
n-wach, line 33: Can you add a comment explaining what this is for?  And include units if this is in seconds/milliseconds/etc.

src/main/webapp/scroll-container.js
n-wach: Why was this change made?  I don't see scroll being called anywhere without an element specified

src/main/webapp/view/transcript/transcript.js
amuamushu: done, thanks!

src/main/webapp/view/transcript/transcript-area.js
```diff
@@ -71,15 +71,18 @@ export default class TranscriptArea {
...
     const transcriptContainer = TranscriptArea.transcriptScrollContainer();
     const ulElement = document.createElement('ul');
     // TODO: Move the class assignment to the HTML.
     ulElement.class = 'mx-auto';
     transcriptContainer.appendChild(ulElement);
     transcriptLines.forEach((transcriptLine) => {
-      ulElement.appendChild(
-          TranscriptLineElement.createTranscriptLineElement(transcriptLine));
+      const transcriptLineElement =
+          TranscriptLineElement.createTranscriptLineElement(transcriptLine);
```
billie-cat, line 82: Does this do any work?

src/main/webapp/view/transcript/transcript-seeker.js
```diff
@@ -114,5 +117,10 @@ export default class TranscriptSeeker {
     return transcriptLinePointer;
   }
 
-  // TODO: Add a method that adds the eventListeners.
+  /**
+   * Returns the eventController.
+   */
```
billie-cat, line 122: This comment isn't useful. I would expand on what the event controller does.

src/main/webapp/view/transcript/transcript-area.js
amuamushu: It calls the constructor for creating a `transcriptLineElement` [link to code](https://github.com/googleinterns/zoomtube/blob/master/src/main/webapp/view/transcript/transcript.js#L46)

src/main/webapp/view/transcript/transcript-seeker.js
amuamushu: gotcha, updated!

src/main/webapp/view/transcript/transcript-area.js
billie-cat: For some reason I thought this was part of the constructor.

Issue #328, opened 2020-08-27
Seek when a discussion comment's timestamp is clicked


n-wach:
Using event controller

PR #339, opened 2020-08-28
Seek when a root comment's header is clicked

n-wach:
Also adds styling to make root comment headers appear clickable.

Closes: #328


src/main/webapp/view/discussion/discussion-comment.js
```diff
@@ -64,25 +64,52 @@ export default class DiscussionComment extends HTMLElement {
...
+    this.setContent();
     this.setTypeTag(comment.type);
   }
 
+  /**
+   * Adds the comment's header to the header slot. For root comments, adds
+   * a click event handler to seek everything to the comment's timestamp.
+   */
+  setHeader() {
+    const headerSpan = document.createElement('span');
```
jaysaleh, line 77: `'span'` is used multiple times, consider making it a class const

n-wach: I think that built-in strings for CSS/events/DOM stuff should remain as literals

src/main/webapp/view/discussion/discussion-area.js
```diff
@@ -72,7 +72,15 @@ export default class DiscussionArea {
   addSeekingListener() {
     this.#eventController.addEventListener((timestampMs) => {
       this.seek(timestampMs);
-    }, 'seek');
+    }, 'seek', 'seekAll');
+  }
+
+  /**
+   * Broadcasts a `seekAll` event for a specific `timestampMs` that was clicked.
```
billie-cat, line 79: This explains what the code does. What does the 'seekAll' broadcast do? I am not sure what this method is supposed to do. So I assume this method is called when the timestamp of a comment is clicked? But that does not explain what the method will do. I would rename the method. Am I understanding correctly?

n-wach: I updated the method name and JSDoc.  Let me know if it's clearer now

src/main/webapp/view/discussion/discussion-area.js
```diff
@@ -72,7 +72,18 @@ export default class DiscussionArea {
...
+    }, 'seek', 'seekAll');
+  }
+
+  /**
+   * Seeks the transcript, discussion, and video to `timestampMs`.
+   *
+   * <p>This is private and should be added as an event listener to every root
+   * discussion header's onclick event.
+   */
+  onTimestampClicked(timestampMs) {
```
billie-cat, line 84: what does onTimestampClicked? Which timestamp was clicked?

n-wach: renamed to `onCommentHeaderClicked`

PR #352, opened 2020-08-28
Adds seeking video function, makes video player private

jaysaleh:
This pr allows seeking the video and makes the video player private.


src/main/webapp/view/video/video.js
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import Synchronizer from '../../synchronizer.js';
+import {millisecondsToSeconds} from '../../timestamps.js';
```
jaysaleh, line 16: These are on separate lines because being on the same line the line width is 81, max linter accepts is 80.

src/main/webapp/view/video/video.js
```diff
@@ -60,14 +62,14 @@ export default class Video {
...
   }
 
-  /** Seeks video to `currentTime`. */
-  seekVideo(timeMs) {
-    // TODO: Removed and implement.
-    console.log('SEEKING VIDEO TO: ' + timeMs);
+  /** Returns current video time of 'videoPlayer' in milliseconds. */
+  getCurrentVideoTimeMs() {
+    return secondsToMilliseconds(this.#videoPlayer.getCurrentTime());
   }
```
billie-cat, line 74: I saw this code in a different PR, do you need to sync?

jaysaleh: Yes syncing now

PR #362, opened 2020-08-31
Create a builder class to store the properties of a transcript language.

amuamushu:
This class is used to store the various properties of a transcript language.

Example XML with the various properties: http://video.google.com/timedtext?v=wXhTHyIgQ_U&type=list

This is the first of 4 pull requests and will only be merged in if the others are approved in time.

1st PR: (This one)  #362
2nd PR: #363
3rd PR: #372
4th PR: #378


src/main/java/com/googleinterns/zoomtube/data/TranscriptLanguage.java
```diff
@@ -0,0 +1,50 @@
...
+@AutoValue
+public abstract class TranscriptLanguage {
+  /** Returns the name of the language. */
+  public abstract String languageName();
+
+  /** Returns the ISO language code for the language. */
+  public abstract String languageCode();
+
+  /** Returns the name of the language in English. */
+  public abstract String languageTranslatedName();
```
billie-cat, line 31: I would rename it as languageNameInEnglish.

languageTranslatedName doesn't imply it will be in english

amuamushu: gotcha, updated!

Issue #335, opened 2020-08-27
Support updating individual comment information when refetching comments


n-wach:
Currently, we only check if a comment hasn't been seen before, but for marking questions as answered, we're gonna also need a way to update an existing comment's type.

PR #367, opened 2020-08-31
Check and update displayed comment type when updating discussion

n-wach:
The update discussion function now cause the comment type to be updated without needing to refresh the page. 

So if you press a button, or someone else presses it, you don't need to refresh the page to see the new type.

Closes: #335


src/main/webapp/view/discussion/discussion-comment.js
```diff
@@ -165,11 +164,15 @@ export default class DiscussionComment extends HTMLElement {
...
-    const typePill = document.createElement('span');
-    typePill.innerText = COMMENT_TYPES[type].name;
-    typePill.classList.add(...COMMENT_TYPES[type].badgeStyles);
-    typePill.slot = DiscussionComment.#SLOT_TYPE_TAG;
-    this.appendChild(typePill);
+    if (this.typePill) {
+      this.typePill.remove();
+    }
+
+    this.typePill = document.createElement('span');
```
billie-cat, line 171: It is a bit weird for the method to be typeTag but the variable to be typePill. I would make them have the same name.

n-wach: changed to typeTag

PR #369, opened 2020-08-31
Parses icon feedback

jaysaleh:
this pull request parses the icon feedback and adds `pared-icon-feedback.js` file which contains a class storing parsed icon feedback data.


src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Parses `iconFeedbackJson` by couting how many times each IconFeedback type
+   * is clicked in each 10 second interval and stores that data in a
+   * ParseIconFeedback object.
+   */
+  parseFeedback(iconFeedbackJson) {
+    let index = 0;
+    let interval = 0;
+    while (index < iconFeedbackJson.length) {
+      let goodCount = 0;
```
n-wach, line 58: I think this could be made a lot simpler if you used a dictionary...

```
const typeCounts = {
GOOD: 0,
BAD: 0,
...
}

while(...) {
typeCounts[iconType]++;
}
...
```

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,100 @@
...
+      let badCount = 0;
+      let tooFastCount = 0;
+      let tooSlowCount = 0;
+      while (index < iconFeedbackJson.length &&
```
n-wach, line 62: Consider refactoring this to avoid nested loops and tracking the index:

```javascript
for (const iconFeedback of iconFeedbackJson) {
  if (timestamp < interval) {
    update interval, append/reset counts
  }
  ... increment count
}
append counts
```

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,100 @@
...
+             iconFeedbackJson[index].timestampMs < interval) {
+        if (iconFeedbackJson[index].type == 'GOOD') {
+          goodCount++;
+        } else if (iconFeedbackJson[index].type == 'BAD') {
+          badCount++;
+        } else if (iconFeedbackJson[index].type == 'TOO_FAST') {
```
n-wach, line 68: Are these constants declared somewhere in the Javascript already?

src/main/webapp/feedback/parsed-icon-feedback.js
```diff
@@ -0,0 +1,47 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/** Stores parsed icon feedback data. */
+export default class ParsedIconFeedback {
+  /* Each index of these represents the number of IconFeedbacks for that
+   * icon type during that 10 seconds interval.
+   */
+  #goodCounts = [];
```
n-wach, line 20: Could you use a dictionary of `type -> array`?

Then you can remove the individual `appendXXCount`s to just take that dictionary of all `typeCounts` and append them all in a `for...in` loop

src/main/webapp/feedback/load-icon-feedback.js
jaysaleh: No

src/main/webapp/feedback/parsed-icon-feedback.js
```diff
@@ -0,0 +1,42 @@
...
+    this.#counts[IconFeedbackUtil.TYPE_GOOD].push(
+        counts[IconFeedbackUtil.TYPE_GOOD]);
+    this.#counts[IconFeedbackUtil.TYPE_BAD].push(
+        counts[IconFeedbackUtil.TYPE_BAD]);
+    this.#counts[IconFeedbackUtil.TYPE_TOO_FAST].push(
+        counts[IconFeedbackUtil.TYPE_TOO_FAST]);
+    this.#counts[IconFeedbackUtil.TYPE_TOO_SLOW].push(
+        counts[IconFeedbackUtil.TYPE_TOO_SLOW]);
+    this.#counts[IconFeedbackUtil.INTERVAL].push(
+        counts[IconFeedbackUtil.INTERVAL]);
```
n-wach, line 40: This can be done with a loop

jaysaleh: still working on this

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,105 @@
...
+             iconFeedbackJson[index].timestampMs < interval) {
+        if (iconFeedbackJson[index].type == IconFeedbackUtil.TYPE_GOOD) {
+          typeCountsAndInterval[IconFeedbackUtil.TYPE_GOOD]++;
+        } else if (iconFeedbackJson[index].type == IconFeedbackUtil.TYPE_BAD) {
+          typeCountsAndInterval[IconFeedbackUtil.TYPE_BAD]++;
+        } else if (
+          iconFeedbackJson[index].type == IconFeedbackUtil.TYPE_TOO_FAST) {
+          typeCountsAndInterval[IconFeedbackUtil.TYPE_TOO_FAST]++;
+        } else {
+          typeCountsAndInterval[IconFeedbackUtil.TYPE_TOO_SLOW]++;
```
n-wach, line 80: replace all of these `if`s with something like:
```javascript
const type = iconFeedbackJson[index].type;
typeCountsAndInterval[type]++;
```

src/main/webapp/feedback/load-icon-feedback.js
jaysaleh: Added util class

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,103 @@
...
+import IconFeedbackUtil from './icon-feedback-util.js';
+import ParsedIconFeedback from './parsed-icon-feedback.js';
+
+/* Handles loading IconFeedback from database and parses data for graph. */
+export default class LoadIconFeedback {
+  static #ENDPOINT_FEEDBACK = '/icon-feedback';
+  static #PARAM_LECTURE_ID = 'lectureId';
+
+  /* Each interval is 10 seconds, used to increment interval. */
+  static #INCREMENT_INTERVAL = 10000;
```
amuamushu, line 25: Can you include the units (ms, s, etc)?

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,103 @@
...
+    this.#lectureId = lectureId;
+  }
+
+  async initialize() {
+    this.#parsedIconFeedback = new ParsedIconFeedback();
+    await this.loadIconFeedbackList();
+  }
+
+  /**
+   * Fetches avaiable IconFeedback from `ENDPOINT_FEEDBACK`
```
amuamushu, line 40: typo
```suggestion
   * Fetches available IconFeedback from `ENDPOINT_FEEDBACK`
```

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,103 @@
...
+    this.parseFeedback(jsonData);
+  }
+
+  /**
+   * Parses `iconFeedbackJson` by couting how many times each IconFeedback type
+   * is clicked in each 10 second interval and stores that data in a
+   * ParseIconFeedback object.
+   */
+  parseFeedback(iconFeedbackJson) {
+    console.log(iconFeedbackJson);
```
amuamushu: can this print statement be removed?

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,103 @@
...
+    let index = 0;
+    let interval = 0;
```
amuamushu, line 59: Calling it an interval seems a bit weird to me considering it's a single number. Maybe call it `intervalLowerBound`?

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,105 @@
...
+
+const PARAM_ID = 'id';
+
+/** Lecture ID stored in `window.location.serach`. */
+const lectureId = getLectureId(window.location.search);
+
+/**
+ * Returns the lecture id from `urlSearchParams`.
+ */
+function getLectureId(urlSearchParams) {
```
amuamushu, line 99: Would it be better to combine this function with the one from `view.js`?

src/main/webapp/feedback/load-icon-feedback.js
jaysaleh: yes im stilling working on this

src/main/webapp/feedback/load-icon-feedback.js
jaysaleh: This file is in a separate path from view as it redirect to another page, this change would require making a parent to both view and feedback, could be done if we have enough time.

src/main/webapp/feedback/parsed-icon-feedback.js
```diff
@@ -0,0 +1,39 @@
...
+    [IconFeedbackUtil.INTERVAL]: [],
+  }
+
+  /**
+     Appends each value in `typeCountsAndInterval` to it's respective
+     dictionary value.
+   */
+  appendTypeCountsAndInterval(typeCountsAndInterval) {
+    for (const type in typeCountsAndInterval) {
+      if (Object.prototype.hasOwnProperty.call(typeCountsAndInterval, type)) {
```
jaysaleh, line 34: https://eslint.org/docs/rules/guard-for-in

jaysaleh: Required for validation checks to pass

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,106 @@
...
+    let intervalLowerBound = 0;
+    let typeCountsAndInterval = this.makeCountDictionary(intervalLowerBound);
+    for (const iconFeedback of iconFeedbackJson) {
+      if (intervalLowerBound < iconFeedback.timestampMs) {
+        this.#parsedIconFeedback.appendTypeCountsAndInterval(
+            typeCountsAndInterval);
+        intervalLowerBound += LoadIconFeedback.#INCREMENT_INTERVAL_MS;
+        typeCountsAndInterval = this.makeCountDictionary(intervalLowerBound);
+      } else {
+        const type = iconFeedback.type;
```
n-wach, line 69: I think this shouldn't be in an `else`.  Cause this way don't you miss counting the first in each interval?

src/main/webapp/feedback/parsed-icon-feedback.js
n-wach: Alternatively you can disable that linter check for this particular block of code

src/main/webapp/feedback/load-icon-feedback.js
jaysaleh: I think this works but I will double check soon

src/main/webapp/feedback/icon-feedback-util.js
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/** Contains conts related to IconFeedback. */
```
amuamushu, line 15: typo?
```suggestion
/** Contains counts related to IconFeedback. */
```

jaysaleh: consts sorry

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,110 @@
...
+
+  #lectureId;
+  #parsedIconFeedback;
+
+  constructor(lectureId) {
+    this.#lectureId = lectureId;
+  }
+
+  async initialize() {
+    this.#parsedIconFeedback = new ParsedIconFeedback();
```
billie-cat, line 35: constructors should not do work, so this could be declared in the constructor.

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,110 @@
...
+    const response = await fetch(url);
+    const jsonData = await response.json();
+    console.log(jsonData);
+    this.parseFeedback(jsonData);
+  }
+
+  /**
+   * Parses `iconFeedbackJson` by couting how many times each IconFeedback type
+   * is clicked in each 10 second interval and stores that data in a
+   * ParseIconFeedback object.
```
billie-cat, line 57: This talks about internal code. A JSDoc, JavaDoc should be high level so that if someone who does not have visibility into the code understands what this does. Imagine if you were reading the JavaDoc for HashMaps(https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html) and it talked about internal code. That would be of no use to you.

src/main/webapp/feedback/load-icon-feedback.js
billie-cat: For stuff like this you should add a TODO.

src/main/webapp/feedback/parsed-icon-feedback.js
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import IconFeedbackUtil from './icon-feedback-util.js';
+
+/** Stores parsed icon feedback data. */
+export default class ParsedIconFeedback {
+  /* Values of each icon feedback type at each interval. */
+  #counts = {
```
billie-cat, line 20: this seems like duplicate code to makeCountDictionary. Can you merge the logic together.

src/main/webapp/feedback/load-icon-feedback.js
jaysaleh: Fixed

src/main/webapp/feedback/parsed-icon-feedback.js
jaysaleh: `counts` is a dictionary mapping `type` -> `array of counts`
`makeCountDictionary` returns a dictionary mapping `type` -> `count`

After parsing each interval, each value in the return of `makeCountDictionary` is appended to the value of the respective `type` in `counts`

src/main/webapp/feedback/load-icon-feedback.js
```diff
@@ -0,0 +1,106 @@
...
+    this.#parsedIconFeedback.appendTypeCountsAndInterval(typeCountsAndInterval);
+  }
+
+  /**
+   * Returns a dictionary mapping each icon type to a value which represents
+   * how many times that icon was clicked in an `intervalLowerBound`.
+   */
+  makeCountDictionary(intervalLowerBound) {
+    return {
+      [IconFeedbackUtil.TYPE_GOOD]: 0,
```
billie-cat, line 82: I just realized how this dictionary is being used. This is not great. When you map X to Y then X is not 2 types. Icons and interval are different data types. I would create a class to wrap this information. The class could be called IntervalIconFeedbackCount (or some other name) and it has a time and a dictionary.

To keep this PR small, please make this change in a second PR which merged into master, then you can sync this PR and use that class to clean this up. Please let me know if you have any questions.

src/main/webapp/feedback/parsed-icon-feedback.js
billie-cat: Please see my previous comment about making this into a class.

billie-cat (changes requested):

src/main/webapp/feedback/parsed-icon-feedback.js
jaysaleh: Fixed

billie-cat (approved):

PR #372, opened 2020-08-31
Fetch the languages from the servlet and display them on the DOM.

amuamushu:
This fetches the list of languages as the user inputs them.

No Selector if user hasn't inputed a url: https://screenshot.googleplex.com/69XQ6NaDe4LrXHE.png
Select if user inputs a valid url: https://screenshot.googleplex.com/5sSJzijXJ6UewLN.png
No Selector if user inputs an invalid url: https://screenshot.googleplex.com/33suqhNR8dNeZgv.png
Message instead of a selector if there is no transcripts available: https://screenshot.googleplex.com/7gmgtbZQv7CQNDW.png

This is the 3rd of 4 pull requests and will only be merged in if the others are approved in time.

1st PR: #362
2nd PR: #363
3rd PR: (This one) #372
4th PR: #378

PR #363, opened 2020-08-31
Create a servlet for fetching the transcript language options.

amuamushu:
This parses the XML for a given video URL containing the various transcript languages and returns it as JSON.

This is the second of 4 pull requests and will only be merged in if the others are approved in time.

1st PR: #362
2nd PR: (This one) #363
3rd PR: #372
4th PR: #378

PR #378, opened 2020-09-01
Parse the transcript with the selected language.

amuamushu:
This updates `TranscriptParser` to parse the transcript in the selected language. 

This is the 4th of 4 pull requests and will only be merged in if the others are approved in time.

1st PR: #362
2nd PR: #363
3rd PR:  #372
4th PR: (This one) #378


src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -95,7 +96,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     String lectureName = request.getParameter(PARAM_NAME);
     Entity lectureEntity = LectureUtil.createEntity(lectureName, videoUrl, videoId.get());
     datastore.put(lectureEntity);
-    initializeTranscript(lectureEntity);
+    String transcriptLanguage = request.getParameter(PARAM_LANGUAGE);
```
n-wach, line 99: Can you add a check for this this to the `validatePostRequest` call?

src/main/webapp/lectures/lectures.js
```diff
@@ -19,6 +19,8 @@ const NO_LANGUAGES_AVAILABLE_MESSAGE =
     'Sorry, there is no transcript available for this lecture.';
 const SELECT_LANGUAGE_OPTION_MESSAGE =
     'Select the language for the transcript.';
+const LANGUAGE_SELECTOR = 'language-input';
```
n-wach, line 22: this is a parameter, right? can u rename it PARAM

src/test/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParserTest.java
```diff
@@ -137,17 +122,40 @@ public void parseAndStoreTranscript_persistDataInDatastoreForShortVideo() throws
...
+    try {
+      TranscriptParser.getParser().parseAndStoreTranscript(
+        LONG_VIDEO_ID, lectureKeyB, /* transcriptLanguage= */ "notAValidLanguage");
+      fail();
+    } catch (IOException e) {
+    }
+  }
+
+  @Test
+  public void parseAndStoreTranscript_noLanguageAvailable_throwsException() throws IOException {
```
n-wach, line 148: can you make these both `throws Exception` b/c they are tests

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -111,13 +113,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
 
   /**
    * Parses and stores the transcript lines in datastore using the {@code lectureKey}
-   * and {@code videoId} properties in {@code lectureEntity}.
+   * and {@code videoId} properties in {@code lectureEntity}. The language for parsing
```
jaysaleh, line 116: I think videoId is interval code here

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
amuamushu: I think it would be better not to for now since the language input is not required. If there isn't a language, the parsing error is caught in the try-catch block.  I updated it to use optional though so `transcriptLanguage` wouldn't be null if a language isn't selected.

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
amuamushu: thanks for catching that, updated!

src/main/webapp/lectures/lectures.js
amuamushu: Sort of, it's the name of selector input tag. It doesn't get added as a parameter to a URL though. I think it'll be more fitting to update it to `INPUT_LANGUAGE`

src/test/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParserTest.java
amuamushu: thanks for catching that, updated!

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
```diff
@@ -96,7 +97,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     Entity lectureEntity = LectureUtil.createEntity(lectureName, videoUrl, videoId.get());
     datastore.put(lectureEntity);
     try {
-      initializeTranscript(lectureEntity);
+      String transcriptLanguage =
+          Optional.ofNullable(request.getParameter(PARAM_LANGUAGE)).orElse("");
+      initializeTranscript(lectureEntity, transcriptLanguage);
```
n-wach, line 102: Is `PARAM_LANGUAGE` required / checked to be non-null in `validatePostRequest`?  That would simplify this

Will the API fail when passed `""`?  If so, I think we should just return an error in validate when it is null

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -113,6 +113,7 @@ public void doPost_invalidLink_badRequest() throws Exception {
   public void doPost_urlAlreadyInDatabase_shouldReturnLecture() throws Exception {
```
n-wach, line 113: is there a test for when the language parameter is missing?

apologies if it was included in a previous PR

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
amuamushu: Yes, it will still fail because no language is selected. I was considering doing that, but we check for errors and return in the beginning of the method. If there's no language, we should still be able to create a lecture.

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
amuamushu: There is one in this PR for `TranscriptParser` but not for the `LectureServlet`. I'll add one here too. Thanks!

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
billie-cat, line 102: Why not pass in Optional<String> for transcript language to initializeTranscript? In that method you can set the empty string. This encapsulates the logic of what to set when there is no language.

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -29,6 +29,7 @@
 import com.googleinterns.zoomtube.data.Lecture;
 import com.googleinterns.zoomtube.utils.LectureUtil;
 import com.ryanharter.auto.value.gson.GenerateTypeAdapter;
+
```
billie-cat, line 32: remove

src/test/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParserTest.java
```diff
@@ -128,19 +129,31 @@ public void parseAndStoreTranscript_persistDataInDatastoreForShortVideo() throws
   public void parseAndStoreTranscript_persistDataInDatastoreForLongVideo() throws Exception {
     Key lectureKeyC = KeyFactory.createKey(LectureUtil.KIND, Long.parseLong(LECTURE_ID_C));
 
-    TranscriptParser.getParser().parseAndStoreTranscript(LONG_VIDEO_ID, lectureKeyC);
+    TranscriptParser.getParser().parseAndStoreTranscript(
+        LONG_VIDEO_ID, lectureKeyC, /* transcriptLanguage= */ "en");
```
billie-cat, line 133: It is good you added /* transcriptLanguage= */. Please add it everywhere though.

billie-cat (changes requested):

f-camp (approved):

src/test/java/com/googleinterns/zoomtube/transcriptParser/TranscriptParserTest.java
amuamushu: done!

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
amuamushu: done!

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -205,4 +210,27 @@ public void getVideoId_shouldFindAllIds() {
...
+  @Test
+  public void doPost_noLanguageParameter_shouldAddToDatabaseAndReturnRedirect() throws Exception{
+    when(request.getParameter(LectureServlet.PARAM_NAME)).thenReturn(TEST_NAME);
+    when(request.getParameter(LectureServlet.PARAM_LINK)).thenReturn(TEST_LINK);
+    
+    servlet.doPost(request, response);
+
+    assertThat(datastoreService.prepare(new Query(LectureUtil.KIND)).countEntities()).isEqualTo(1);
+    verify(response).sendRedirect("/view/?id=1");
```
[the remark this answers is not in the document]
amuamushu: I had that there because I'm checking that a lecture is still successfully created even without having a transcript language parameter value.

src/main/java/com/googleinterns/zoomtube/servlets/LectureServlet.java
amuamushu: Gotcha, I originally figured it would be nice to extract the value as early as possible, but encapsulating it makes sense. Updated!

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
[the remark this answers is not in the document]
billie-cat: I was wondering if this was a static string and if you should use the variable name. This is mostly a nit, you do not need to change it if it doesn't make sense.

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
```diff
@@ -137,6 +143,8 @@ public void doPost_videoHasNoTranscript_shouldReturnLectureAnyways() throws Exce
   public void doPost_urlNotInDatabase_shouldAddToDatabaseAndReturnRedirect() throws Exception {
     when(request.getParameter(LectureServlet.PARAM_LINK)).thenReturn(TEST_LINK);
     when(request.getParameter(LectureServlet.PARAM_NAME)).thenReturn(TEST_NAME);
+    when(request.getParameter(LectureServlet.PARAM_LANGUAGE))
+        .thenReturn(/* transcriptLanguage= */ "en");
```
billie-cat, line 147: the string in /* =*/ should be the argument name. I assume thenReturn does not have a variable name of transcriptLanguage. For thenReturn it is obvious what the method does, so you should omit the /* =*/ comment. When I said to add this I meant for the parseAndStoreTranscript methods where it isn't obvious what the string might represent.

billie-cat (changes requested):

billie-cat (approved):

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
amuamushu: gotcha, updated.

src/test/java/com/googleinterns/zoomtube/servlets/LectureServletTest.java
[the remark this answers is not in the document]
amuamushu: Gotcha, I think I'll leave it as is then because something similar was done for the other tests.

PR #389, opened 2020-09-01
Fix issue where the currentElement is not defined in scrollContainer

amuamushu:
If auto scroll is started before the video starts playing, we would have trouble calling `scrollToTopOfContainer` because `currentElement` is undefined. This fixes the issue by setting a default value for this situation.


src/main/webapp/scroll-container.js
```diff
@@ -77,8 +77,12 @@ export class ScrollContainer extends HTMLDivElement {
    * If automatic scrolling is enabled or `forceScroll` is true, then scrolls
    * the container such that `element` is at the top of the container.
    * Otherwise, does nothing.
+   *
+   * <p>If `element` is undefined, the element to scroll is defaulted to the
+   * first element in the container storing the scrollable elements.
    */
-  scrollToTopOfContainer(element, forceScroll = false) {
+  scrollToTopOfContainer(
+      element = this.lastChild.firstChild, forceScroll = false) {
```
amuamushu, line 85: `this.lastChild` refers to the container in ScrollContainer that store the scrollable elements (transcriptLineElements or comments)

billie-cat: Please add what you wrote as a comment in the code.

f-camp (approved):

src/main/webapp/scroll-container.js
amuamushu: I updated the method header to mention this in the second paragraph.

PR #391, opened 2020-09-01
Update storage setting for the local datastore test configuration.

amuamushu:
There was no storage setting for the datastore, which caused one of the tests to fail.


n-wach:
Is this done in all other tests? Should it be done?

jaysaleh:
> Is this done in all other tests? Should it be done in them?

It is only done in `TranscriptServletTest.java`

n-wach:
> > Is this done in all other tests? Should it be done in them?
> 
> It is only done in `TranscriptServletTest.java`

Can we expand this PR to add this to every other test?  Because it seems like this is a good thing to do to speed up the tests

amuamushu:
> > > Is this done in all other tests? Should it be done in them?
> > 
> > 
> > It is only done in `TranscriptServletTest.java`
> 
> Can we expand this PR to add this to every other test? Because it seems like this is a good thing to do to speed up the tests

done!

f-camp (approved):

jaysaleh (approved):

